PR #4, opened 2015-05-15
feat: router

gdi2290:
There is a problem with the router not recognizing existing uri


gdi2290:
added to master https://github.com/angular-class/angular2-webpack-starter/commit/7b23e35c15ea21e6c56560be621b4f9c62ea35dd

PR #18, opened 2015-06-01
Update webpack.config.js

EladRK:
typo

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -67,7 +67,7 @@ module.exports = {
       '.webpack.js',
       '.web.js'
     ],
-    // Todo: learn more about aslias
+    // Todo: learn more about alias
     alias: {
       // When Angular2 has a TypeScript build
       // 'angular2': grootNode('angular2/es6/prod'),


gdi2290:
thanks! haha if you could help me clean up and make the examples more clear that would be awesome

EladRK:
We're now working on porting our knockout.js based website to angular 2, and as we're using your starter kit i'm pretty sure we'll stumble upon many things. I'm just glad I can give a hand back :)

gdi2290:
@EladRK haha thanks but keep in mind angular2 is very early so it's not production ready yet. I would only suggest remaking it as a simple example

EladRK:
shhhhh.. i just managed to convince upper management that it's safe enough - don't let them know!

we're starting a very needed refactoring and it'll take us no less than 3 months to merge this branch to production (hopefully). we're assuming that 3 months from now it'll be stable enough. for what we're doing now it's still unclear if we can actually working with it, but so far we haven't stumbled upon any showstoppers

PR #53, opened 2015-08-13
PhantomJS compatible unit tests

tamascsaba:
Fix issue: https://github.com/angular-class/angular2-webpack-starter/issues/45
It is only workaround not best solution.

```
{ test: /reflect-metadata/, loader: "imports?require=>false" },
```
Because reflect-metadata require('crypto') (node-libs-browser/browserify-crypto) module which not compatibile with PhantomJS


gdi2290:
@tamascsaba how is es6-shim compared to traceur-runtime? does it include more polyfills?

tamascsaba:
With traceur, Phantomjs is crashed when I add Function.prototype.bind pollyfill and the crash log is not readable :(
 
```
ERROR [launcher]: PhantomJS crashed.
```

tamascsaba:
But I think Traceur is more compact pollyfill, es6-shim is little bit heavy.
I play with [autopolyfiller](https://github.com/azproduction/autopolyfiller) and [autopolyfiller-loader] (https://github.com/deepsweet/autopolyfiller-loader) but more time need for good solution.

gdi2290:
merged, thanks!

PR #72, opened 2015-09-11
use rimraf for better windows support

robrich:


gdi2290:
thanks!

gdi2290:
I really should just use gulp tbh

PR #78, opened 2015-09-20
Update tsconfig.json for Typescript 1.6

jimthedev:
Update the version of the tsconfig.json to match the TypeScript version
in package.json and fixes the path to lib.dom.d.ts which moved as of TS
version 1.6

This fixes annoying red squiggly line errors in visual studio code (and potentially
other editors) relating to module, experimentalDecorators or emitDecoratorMetadata
not being set.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.5.0",
+  "version": "1.6.0",
   "compilerOptions": {
     "target": "es5",
     "module": "commonjs",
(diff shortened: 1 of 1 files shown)


jimthedev:
This is all set. /cc @gdi2290

gdi2290:
thanks!

gdi2290:
(just found out the version number was only added for atom support and the tsconfig spec has no mention of a version field)

jimthedev:
Weird. I saw it in there but no one I talked to had heard of it. Does atom still require it?

jimthedev:
I do wish Visual Studio Code would not silently ignore tsconfig.json when a file path is wrong in the 'files' array. Seems like skipping that file might be more sensible.

PR #79, opened 2015-09-22
Fix Protractor/Selenium Configuration

dotcs:
To get e2e tests running, I had to fix some issues with Protractor/Selenium. First of all I had to adjust the port of the `baseUrl` to `3000` since the webpack dev server is running on this port. 

Then I ran into an issue in which it was impossible to get the WebDriver session started. This is documented [here](https://github.com/angular/angular-seed/issues/251). The [solution](https://github.com/angular/angular-seed/issues/251#issuecomment-129217595) by @josefsalyer worked for me, so I suggest to use it in this repo as well.

### DIFF
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -1,7 +1,7 @@
 // @AngularClass
 
 exports.config = {
-  baseUrl: 'http://localhost:8080/',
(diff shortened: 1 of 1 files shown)


gdi2290:
thanks!

PR #82, opened 2015-09-25
Fix: x-large component style

### LINKED ISSUE #81: Directive provided in the app.ts does not work
element.nativeElement.style.fontSize = 'x-large'; instead of element.nativeElement.fontSize = 'x-large';

E.g. missing style in the example provided.

dotcs:
Fixes #81 and makes use of the component in the footer to better demonstrate its usage.


gdi2290:
thanks!

PR #87, opened 2015-09-27
Update ac-autosuggest.ts

jennyraj:
I think this is what meant ..


jennyraj:
Sorry , the constant GITHUB_BINDINGS  does not seem to work. Leaving it as it was...i.e. GithubService

PR #88, opened 2015-10-01
Add typedoc support

asgerjensen:
This PR adds typedoc support for the src/* folder.

It does spit out some errors due to duplicate symbols on some of the angular2 imports (Promise amongst others), but it will parse and process the src folder just fine.


gdi2290:
thanks!

PR #95, opened 2015-10-14
Update protractor package version as the current 2.1.0 fails on node 4.1.1+

krigton:
Update protractor package version as the current 2.1.0 fails on node 4.1.1+


gdi2290:
thanks!

krigton:
welcome!

PR #96, opened 2015-10-15
Update app.ts

incrediblesound:
Update Directive description grammar:
ORGINAL "XLarge is a simple directive to show how one of made"
NEW " XLarge is a simple directive to show how one is made"

Changed "of" to "is"


gdi2290:
seems legit

PR #100, opened 2015-10-18
move the static file serving as the latest fallback. the /api request…

krigton:
… won't work without this change


gdi2290:
thanks!

krigton:
welcome.

Sent from Yahoo Mail on Android

PR #103, opened 2015-10-21
Refactor for correct reference to Rx module

chriscurnow:
Original examples had
```javascript
import * as Rx from 'rx'
```

This should have been

```javascript
import * as Rx from '@reactivex/rxjs'
```


gdi2290:
thanks!

PR #105, opened 2015-10-22
Clarified some comments re starting server.

chriscurnow:
The comment 

```javascript
"You must run these commands for the Http API to work in another process "
```
is confusing. On initial reading it sounds like we want the Http API to run in another process.

Whereas in fact we want these commands to run in another process. So changed to 

```javascript
"You must run these commands in another process for the Http API to work"
```

Also made this clearer in the comments just under onInit().


gdi2290:
thanks!

PR #111, opened 2015-10-26
Refactor to fix broken links in large-app

chriscurnow:
Went through large-app to the point where it would compile and run to display main banner.
Still some things not working – see commented out lines in large-app/bootsrap.ts


gdi2290:
thanks!

PR #113, opened 2015-10-27
fix(draggable.ts): Use toRx() of EventEmitter

SekibOmazic:
Use public method toRx() of EventEmitter insted of _subject.

### DIFF
--- a/examples/rx-draggable/directives/draggable.ts
+++ b/examples/rx-draggable/directives/draggable.ts
@@ -27,18 +27,18 @@ export class Draggable {
     this.element.nativeElement.style.position = 'relative';
     this.element.nativeElement.style.cursor = 'pointer';
 
-    this.mousedrag = this.mousedown._subject.map(event => {
+    this.mousedrag = this.mousedown.toRx().map(event => {
         event.preventDefault();
         return {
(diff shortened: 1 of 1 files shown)


SekibOmazic:
@gdi2290 Maybe you could help. Mouse pointer jumps 8px up and 8px left (margin of body tag) when start dragging and then moves smoothly. Any clue why?

dotcs:
@SekibOmazic The 8px offset comes from the browser's default margin of the body. If you set `body { margin: 0 }`, the problem is solved. We could fix it by removing the browser's default margin or perhaps it would be better to make use of [normalize.css](https://necolas.github.io/normalize.css/) or suchlike.

SekibOmazic:
@dotcs Thanks. The `body { margin: 0 }` solves the problem. But I hoped it could be done without using/changing css.

SekibOmazic:
@dotcs Sorry for bothering, but please take look at this example:

http://fiddle.jshell.net/mattpodwysocki/pfCqq/show/

It has `body {margin:0}` (loaded with bootstrap.min.css) but if you disable it and try to drag picture it works smoothly. You can even change margin to `body{margin 50px}` it still goes smoothly.

PR #118, opened 2015-10-28
Fixed large-app example paths

zarkosusnjar:
Paths to other examples were missing ../ and examples depending on location.


dotcs:
For the file `examples/large-app/components/app.ts` it is the same if you do `../../simple-component/home/home` or `../../**../examples**/simple-component/home/home`, because the file already is located in the `examples` folder.

zarkosusnjar:
I understood that in order to test some example I need to paste it to src directory. If I do that large-app references other examples. Maybe README needs some update, I see on gitter many people don't get this copy paste thing.

dotcs:
To run an example, you don't need to copy anything. It's enough to tell webpack that the `bootstrap.ts` of the example should be used instead of the one in the `src` folder.

Let's assume you want to run the 'rx-autosuggest' example. Change [these two lines](https://github.com/angular-class/angular2-webpack-starter/blob/master/webpack.config.js#L78..L80) in the webpack config file by removing the comment signs in line 78 and comment out line 80. Then restart webpack and you're done.

I'll have a look the README to see how we can better describe the procedure there. Thanks for your feedback!

dotcs:
@gdi2290 What is the recommended approach to use the examples? If the idea is to copy&paste the code to the `src` folder, @zarkosusnjar's PR should be accepted (and we should check all other imports in the examples, too). I thought that the main idea is to switch examples via webpack config as described in my post above. But reading example's README sounds like the recommended approach is more or less to copy&paste the code ...

We could also use webpack to resolve the examples automatically by adding the examples folder to the resolved variables like it's done [here](https://github.com/angular-class/angular2-webpack-starter/blob/master/webpack.config.js#L97) for 'rx'. Then we could always write `import {Home} from 'examples/simple-component/home/home';` regardless of where the file that needs the import is located at. It's a matter of taste, so which way do you prefer?

SekibOmazic:
Switching via webpack config is my preferred solution. IMHO it's more convenient. I'd go that way.

@gdi2290 ?

gdi2290:
sorry I can't accept this. At the moment you would copy/paste or change the entry in webpack.config to quickly view the example

PR #83, opened 2015-09-25
Enhancement: Include Bootstrap 3.3.5 (CSS + JS)

dotcs:
This PR makes use of Bootstrap 3.3.5 and jQuery 2.1.4. It demonstrates the usage of the tooltip plugin. I would suggest that you don't merge this to master, but make a separate branch that includes Bootstrap.

The PR should fix #16 and #61.

Please note, that it was necessary to include a modified version of the jQuery description (`jquery.d.ts`) to make this work with the rest of the setup. The fix has been provided by [@chaosmail](https://github.com/borisyankov/DefinitelyTyped/issues/2734#issuecomment-118950111). Details on this topic can be found there.

PR #127, opened 2015-11-08
ng2-bootstrap + example

dotcs:
This PR adds `ng2-bootstrap` (as requested in https://github.com/AngularClass/angular2-webpack-starter/pull/83). `ng2-bootstrap` provides the ng2 components, `bootstrap` is purely used for the CSS file. The latter could be replaced with a CDN version, but I prefer to be able to work offline which is why I added also the `bootstrap` dependency.

To make things work, I had to [remove the `/node_modules/` regex](https://github.com/dotcs/angular2-webpack-starter/commit/f6a520138588300f27d6b1042f3d534117c23749#diff-11e9f7f953edc64ba14b0cc350ae7b9dL131) in webpack's typescript module loader definition. I tried to only include the `ng2-bootstrap` module - without success. The regex `node_modules\/((?!(ng2-bootstrap)).)*` did not work, although it should exclude all but the `ng2-bootstrap` folder. If the exclusion of `node_modules` is necessary for any other parts of this project, we should find a way around this issue before merging.

/cc: @gdi2290

### DIFF
--- a/examples/ng2-bootstrap/bootstrap.ts
+++ b/examples/ng2-bootstrap/bootstrap.ts
@@ -1,0 +1,7 @@
+/// <reference path="../typings/_custom.d.ts" />
+
+import {bootstrap} from 'angular2/angular2';
+
+import {App} from './components/app';
+
+bootstrap(App);
--- a/examples/ng2-bootstrap/components/app.ts
+++ b/examples/ng2-bootstrap/components/app.ts
@@ -1,0 +1,23 @@
+/// <reference path="../../typings/_custom.d.ts" />
+/// <reference path="../../../node_modules/ng2-bootstrap/ng2-bootstrap.d.ts" />
+
+require('!!style!css!bootstrap/dist/css/bootstrap.css');
+
+import {Component, View} from 'angular2/angular2';
+import {RouteConfig} from 'angular2/router';
+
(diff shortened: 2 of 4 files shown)


gdi2290:
exclusion of `node_modules` is for performance reasons. perhaps @valorkin (creator of ng2-bootstrap) can provide some insight otherwise LGTM

dotcs:
@gdi2290 Performance matters, you're absolutely right. I have had another look at the regex and found out how to tweak it. The regex does now exclude all but the `ng2-bootstrap` package. Should be fine to merge now. Please have another look at it.

dotcs:
@gdi2290 The strategy is now to keep examples separated, right? Shall I port this PR to the new repo that contains the examples? https://github.com/AngularClass/angular2-examples

gdi2290:
can you do that? I'll add you to that repo

lanocturne:
@dotcs Thanks for the PR example. I am getting font 404 errors when applying glyph-icons:  
```bash
GET http://localhost:3000/448c34a56d699c29117adc64c43affeb.woff2 
(index):1 GET http://localhost:3000/fa2772327f55d8198301fdb8bcfc8158.woff 
(index):1 GET http://localhost:3000/e18bbf611f2a2e43afc071aa2f4e1512.ttf 
```

PR #128, opened 2015-11-11
Add some simple tests to demonstrate testing directives and components

dotcs:
I have followed the discussion in https://github.com/AngularClass/angular2-webpack-starter/issues/124 and got tests up and running. I have added some simple tests for directives and components. This should be a good starting point for adding more tests that demonstrate also tests of other facets of ng2.

PR #129, opened 2015-11-11
Move browser dom adapter setup to spec.bundle.js

dotcs:
See https://github.com/AngularClass/angular2-webpack-starter/pull/128#discussion-diff-44580177


gdi2290:
thanks!

PR #138, opened 2015-12-04
chore(Cleanup): removed unused 'View' and using the recently renamed ngOnInit now with alpha.47

NathanWalker:
Also with recent builds, we no longer need to specify `CORE_DIRECTIVES` or `FORM_DIRECTIVES` as those come out of the box now.


gdi2290:
thanks!

PR #145, opened 2015-12-10
Fix : polyfill added for phantomjs as needed for karma tests.

mcha:
Hi guys, 

Karma test just can't work as is. 

I suggest this quick fix to ensure they can be launched.

Best regards,
Michel.


gdi2290:
thanks!

PR #150, opened 2015-12-13
spec-bundle.js now requires core-js instead of es6-shim

jhuntoo:
Unit tests did not run after a fresh clone.

spec-bundle.js was requiring 'es6-shim', updating to core-js appears to have solved this for me.


gdi2290:
thanks!

PR #157, opened 2015-12-15
update deprecated imports to 2.0.0-alpha.53

loose11:
Due the update from version 2.0.0-alpha.52 to 2.0.0-alpha.53 there are deprecated versions of imports. I added the new imports.


gdi2290:
thanks!

PR #166, opened 2015-12-18
Remove global dependencies

umayr:
There is no need to install any package that is mentioned in the [README.md](https://github.com/AngularClass/angular2-webpack-starter#dependencies) to be installed globally. We can use their local versions.

Cheers.


gdi2290:
npm normally checks local first but for some reason developers still run into problems which is why I asked everyone to still everything globally. we would have to prepend each call with `node` for example
`node ./node_modules/.bin/webpack` for windows users unless I'm mistaken

mohsen1:
You don't need to advise for any global module installation [npm scripts](https://docs.npmjs.com/misc/scripts) will automatically use executables from `node_modules/.bin`. A script like [this](https://github.com/mohsen1/angular-json-schema-form/blob/master/package.json#L11) will work even if you don't have Gulp installed globally. This also works on Windows.

I wrote about it a little while ago http://azimi.me/2014/12/18/npm-start.html

gdi2290:
thanks for the PR but it's unnecessary as mentioned above

PR #168, opened 2015-12-19
Make `npm start` install the dependencies

mohsen1:
`typings install` is setup to be a post install script so this will
install all the dependenies


gdi2290:
I originally had this but it turned out to be too much magic for some developers and is also a problem with ci environments

asafyish:
I don't like it either.

gdi2290:
I'm going to close this but if you like you can make a section in the read called `extending` or something for how people can expand upon the starter

PR #181, opened 2015-12-26
Add starter repo

antonybudianto:
See the starter [here](https://github.com/antonybudianto/angular2-starter)


gdi2290:
I removed the links but I also plan on either renaming this repo or making two more with different setups

PR #195, opened 2016-01-04
Update app.ts

vladotesanovic:
If someone try to navigate on non existent route, redirect him to Index route.


gdi2290:
nice! :+1:  thanks!

Issue #185, opened 2015-12-28
Naming conventions + style guide and best practices


manavsehgal:
In the spirit of following best practices for this starter. Any thoughts on naming conventions and style guide for Angular2, that this starter can follow? 

Kind of like Angular2 version of the famous [johnpapa angular 1 style guide](https://github.com/johnpapa/angular-styleguide). I am sure many of the recommendations can be reused for Angular 2 with few changes.

Currently I notice following issue with component naming:

Angular 2 official docs seem to suggest using `app.component.ts` and `AppComponent` naming convention. More verbose. 

PS: I have not looked up the design docs of Angular2 if there are "official" docs that can kickstart such a style guide.

gdi2290:
yeah good idea I can start one

manavsehgal:
Great! I will be happy to contribute.

martinoppitz:
@gdi2290 assign this ticket to me.

dsebastien:
At work we've just started investing in AngularJS (time pressure), but since Angular 2 is just around the corner, we're trying hard to make things easier when we decide to upgrade/migrate.

As far as I can see for now, I think that AngularJS 1.5 apps written in TypeScript can share a (partly) common style guide (e.g., folder structure, folder-per-feature idea, focus on componentization, usage of classes, ...

If you're interested, I'll document here the things that we come up with along our project.

gdi2290:
@dsebastien I'm interested if you can that would help everyone in the Angular community

manavsehgal:
@dsebastien @gdi2290 Let us start a wiki page for documenting the naming conventions. We can use this thread to report issues or debate alternatives. 

Agree with @dsebastien that we can reference existing conventions so it is easy for folks migrating over from Angular1. 

Where missing, we can also lookup other best practices - like [API naming conventions from Apigee](https://pages.apigee.com/rs/apigee/images/api-design-ebook-2012-03.pdf) could possibly help navigation / router related naming conventions.

gdi2290:
I created a wiki page here https://github.com/AngularClass/angular2-webpack-starter/wiki/Best-Practices I'll update next week so if you guys can update it before then it would be awesome to see everyone opinion first

choopage:
I added my opinion to the wiki page you created @gdi2290

manavsehgal:
Updated best practices.

MrLehiste:
[...]

PR #196, opened 2016-01-04
chore(Conventions): naming standards

NathanWalker:
closes https://github.com/AngularClass/angular2-webpack-starter/issues/185

I'm not saying merge this actually, however this is how the naming conventions would play out here.

I've had mixed feelings about them for awhile but in general, like them.
I've just never been a big fan of long filenames which is why I'm not in love with `whatever.component.ts` vs. `whatever.ts`, however it does make things clear.
This becomes more idiomatic when you consider most conventions suggest placing files inside a folder called `components` such as shown here in this PR.
So when opening files and you see:
`app/components/home/home.component.ts`
vs.
`app/home/home.ts`

First, for large projects with more than 2-3 nested hierarchies, you would need a screen or IDE at least wide enough to show the long path and filename when you have multiple tabs open to discern the difference between 2 of the same named files in different folders, ie.:
`app/component/home/home.component.ts`
vs.
`app/component/users/profile/home.component.ts`
^ just an example but it can definitely happen ;)

But in the end, I digress. Whatever the community believes is universally great.

If this is the way to go, then by all means, merge :+1:


src/app/components/home/home.component.ts
```diff
@@ -27,7 +27,7 @@ import {XLarge} from './directives/x-large';
   // Every Angular template is first compiled by the browser before Angular runs it's compiler
   template: require('./home.html')
 })
-export class Home {
+export class HomeCmp implements OnInit {
```
NathanWalker: Even though using `ngOnInit` will work without implementing `OnInit`, it appears to be recommended to still implement for TypeScript sake:
https://github.com/angular/angular/issues/5814

gdi2290: I didn't include this on purpose

src/app/components/app.component.ts
```diff
@@ -36,10 +36,10 @@ import {Home} from './home/home';
   `
 })
 @RouteConfig([
-  { path: '/', component: Home, name: 'Index' },
-  { path: '/home', component: Home, name: 'Home' }
+  { path: '/', component: HomeCmp, name: 'Index' },
+  { path: '/home', component: HomeCmp, name: 'Home' }
 ])
-export class App {
+export class AppCmp {
```
gdi2290: if it's `.component` then the name should also be `AppComponent`

gdi2290:
I have my own reasons for the current structure that I will be covering in a blog post in the future that may conflict with the current naming conventions

package.json
```diff
@@ -70,7 +70,7 @@
     "tslint-loader": "^2.1.0",
     "typedoc": "^0.3.12",
     "typescript": "^1.7.3",
-    "typings": "^0.3.1",
+    "typings": "^0.4.1",
```
gdi2290: can you make a separate pull-request for this?

src/app/components/app.component.ts
NathanWalker: Certainly is the going standard out there.
I guess this PR is almost more for discussion :)

I just shutter at the thought of naming all components with full suffix `Component`.... just for simplicity sake. When shorthand is ample, why not I guess.

gdi2290: if `Cmp` then `.cmp.ts`

NathanWalker:
@gdi2290 Absolutely! Don't merge (as mentioned)! However, maybe we can keep the PR open just to see if others have thoughts that may help form the standards or reach consensus?

Idk, I wish I didn't have mixed feelings about the going conventions/standards, but it seems they are amiss on conveniences and simple shorthand.

package.json
NathanWalker: Will do!

src/app/components/app.component.ts
NathanWalker: ^ I like that. At end of day, I like to use what is universally found in the community. It just makes bringing new people into a project easier because they see what they are used to and have seen everywhere else.

So I like that personally, but don't want to use it if my project would be the only one using it ;)

gdi2290:
I wish there was a convert to issue button. I'll include this in the FAQ

PR #197, opened 2016-01-04
chore(Typings): updated dev to 0.4.1

NathanWalker:


gdi2290:
thanks!

NathanWalker:
btw, love the new refactorings, nice work as always!!
Production build is awesome with configurable setup in vendor. Very similar to what I've ended up doing in other projects.

gdi2290:
thanks, I'm planning on adding gulp and renaming the repo ;)

NathanWalker:
oh wow, to do some extra handling of the build? or to replace webpack?

gdi2290:
gulp with webpack as the bundler. without gulp things like serviceworker becomes more of a problem and gulp with webpack seems to be easier for people to grasp

gdi2290:
I might end up making it another repo and stripping this one to suggest the other one for more

gdi2290:
the problem is when you introduce development/production/test build the complexity sky rockets

NathanWalker:
Sounds excellent.

serviceworker is so exciting. I wish there were a way to implement with ng2 without having to work in any ng2 specific parts as seen here: https://github.com/robwormald/ng2-application-shell .. but I don't think there is.

I like the thought of the 2 repos. I think I'm with you on thinking that a lot of company's and teams need a simple starter to pull off quick prototyping whereas having a separate more sophisticated and complete setup for large production apps is very appreciated and valuable to the community at large. Thanks for all you're doing here, it's incredibly helpful.

gdi2290:
thanks. It's interesting to see that you rather not have ng in your service worker but that does make sense. The original plan was fullstack repo and the webpack repo so the webpack repo will be great for getting started and prototyping while the fullstack one will provide you with everything to go into production

PR #237, opened 2016-01-19
feat(webpack): add compression plugin

### LINKED ISSUE #234: Add CompressionPlugin
What do you think about adding [CompressionPlugin](https://github.com/webpack/compression-webpack-plugin)? I usually statically compress the build assets during the build rather than dynamically on a Web server, it has a couple of advantages:

1. Saves computation resources on the server
2. Simplifies a bit the Web server configuration - no need to configure threshold, ratio, buffers (nginx-specific varies on the platform), the file types that should be compressed.
3. Allows to calculate the real application payload size after the build and report/draw charts on CI.

antonmoiseev:
* Compresses `*.css`, `*.html`, `*.js` and `*.map` build assets
* Uses gzip with maximum compression level - 9

Closes #234

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "bluebird": "*",
     "blueimp-tmpl": "*",
     "colors": "^1.1.2",
+    "compression-webpack-plugin": "^0.2.0",
     "copy-webpack-plugin": "^0.3.3",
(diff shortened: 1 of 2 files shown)


gdi2290:
thanks!

Issue #262, opened 2016-01-27
mangle breaks prod build


mastertinner:
As long as I leave `mangle: { screw_ie8 : true }` as it is set in `webpack.prod.config.js`, my `npm run build:prod` doesn't work. It builds fine but when I try to open the website in my browser (Chrome), I'm getting the following error: `EXCEPTION: No Directive annotation found on t`

Setting `mangle: false` fixes the problem.

gdi2290:
this is a problem with Angular 2

mastertinner:
Thanks, but then shouldn't we disable mangling until beta.2 is out?

gdi2290:
@mastertinner can you make a pull-request and comment about it in prod webpack configs

mastertinner:
Apparently this problem persists in beta.2...

gdi2290:
yup there's still a problem with RouterLink when minified

cristi-badila:
I stumbled into a similar issue while running an app I'm working on in production mode. When using the mangle names option, the outputed javascript results in angular generating a recursive `viewFactory` that will eventually throw a `Maximum call stack size exceeded` exception. I isolated the issue in a repository [here](https://github.com/cristi-badila/angular-mangle-issue/tree/master) which contains both the working non mangled js and the mangled js which results in angular throwing exceptions. Hope this helps.

gdi2290:
I found a workaround via https://github.com/AngularClass/angular2-webpack-starter/commit/1a4edb9693d60ae3b39810bb75b4199123b798ad

yfain:
The workaround suggested by @gdi2290 worked in a couple of betas, but now the mangle:false seem to be the only option (I'm using Webpack with Beta.14).

PR #265, opened 2016-01-27
disable mangle for now

mastertinner:
mangling is broken in Angular 2 beta.1 as described in #262. That's why it should be disabled until Angular 2 beta.2 is out.


gdi2290:
thanks!

PR #292, opened 2016-02-02
Optimize console output for testing

cschroeter:
To keep track of larger projects, I would like to suggest to use the karma-mocha-reporter and summarize the coverage report. 
If you want to have a look at the coverage per file, the HTML output is much more convenient then displaying the line numbers in the console.

**Before:**
before
**After:**
after


gdi2290:
thanks!

PR #301, opened 2016-02-06
fix(typings.json): Wrong zone url

alfonso-presa:
Zone url is incorrect.


gdi2290:
thanks!

alfonso-presa:
Thanks to you. Keep it up!

PR #320, opened 2016-02-11
add hot module replacement for dev-server Fixes #319

### LINKED ISSUE #319: Webpack Dev Server Hot Module Reload
Why not using the hot reloader from the dev server?

NehrDani:
This adds live reload to the dev-server.


NehrDani:
hmh hot reload with typescript isn't that easy.

every bundle file needs...

```
if (module.hot) {
  module.hot.accept();
}
```

... in it's file, to be accepted as hot module.

Problem is that the transpiler won't like it, because 'module' is not defined. Its a variable only available in the dev-server-environment.

you could declare it like 
```
declare let module: any;
```

I'll try to make another pull request with it.

NehrDani:
This should do the trick.
Perhaps there is a smarter solution, but it works now.

The issue that you mustn't bootstrap after DOMContentLoaded for it to work bothers me.

I don't use bootstrap inside a event listener in my own projects, so this issue nevers appeared to me.

gdi2290:
thanks!

PR #330, opened 2016-02-13
Removed typescript typings install from quick start - Fixes #309

### LINKED ISSUE #309: Update Quick Start section
Please remove this step from the Quick Start section
```bash
# install TypeScript typings
npm run typings-install
```
There is already a postinstall script in package.json that will install the typings.

rjmreis:
Removed the instruction to install typescript typings as there is already a postinstall npm script that will do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,6 @@ cd angular2-webpack-starter
 # install the repo with npm
 npm install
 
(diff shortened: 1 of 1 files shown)


gdi2290:
thanks!

PR #348, opened 2016-02-22
Webpack configuration more DRY between production and dev

Spittal:
Made a more generic `webpack.make.config.js` file that includes all the instructions on how to build the project with both production and development. The `webpack.config.js` and the `webpack.prod.config.js` files are now requiring the new make file and passing in some arguments (in this case just an option called `PRODUCTION`). Based on these arguments the make file will change how it returns the config object.

Using this method you don't need to edit two files when you add a loader, or change a plugin. They are using the same configuration file and that makes it much more DRY!


gdi2290:
the goal was to make webpack approachable and easier to reason about what's going on rather than dry. this pull-request introduces too much magic for developers starting out with webpack so I cannot merge it even though it's better for developers who already know webpack

samvloeberghs:
I back this comment from @gdi2290 and wanted to make the same comment before.

Although I like the idea of configuration and the DRY principle, separating the config files gives you more flexibility and readability, with the little compromise of having to repeat yourself. 

This also gives you more options to do different stuff depending on your environment. Putting too much logic in one file will also bloat it a lot!

Spittal:
Totally fair!

There should however be some documentation either in the wiki or README that mentions you need to edit both configurations if you're adding loaders like scss or postcss.

gdi2290:
@Spittal great idea! can you edit this page https://github.com/AngularClass/angular2-webpack-starter/wiki/How-to-DRY-up-your-webpack-config I can also change the title or anything

PR #357, opened 2016-02-26
Fix directive for route active

ggaulard:
Subscribe to the QueryList instead of the Router


gdi2290:
what does it fix?

ggaulard:
I've create a new PR with plunkr preview of the fix.

PR #359, opened 2016-02-26
Fix: Add bail option to build:prod so that CI builds fail when they should

jhuntoo:
My CI build just deployed a mangled version of angular2-webpack-starter because `npm run build:prod` exits with 0 when a loader is missing in node_modules.

This PR adds the --bail option which will cause the webpack build process to exit early, and importantly exit with a non-zero status.


package.json
```diff
@@ -18,7 +18,7 @@
     "watch:prod": "webpack --watch --config webpack.prod.config.js --progress --profile --colors --display-error-details --display-cached",
     "build": "npm run build:dev",
     "prebuild:dev": "npm run clean:dist",
-    "build:dev": "webpack --progress --profile --colors --display-error-details --display-cached",
+    "build:dev": "webpack --progress --profile --colors --display-error-details --display-cached --bail",
```
samvloeberghs: better to put it on the build:prod task than :p

jhuntoo:
@samvloeberghs Derp...  thanks, made the fix in different branch!

samvloeberghs:
and now squash :p

gdi2290:
you should squash it but I'll just merge it. thanks!

PR #368, opened 2016-03-01
Fix directive for route active

ggaulard:
This fix the directive that put a 'active' class on active link.
In a case with a 2 level of routeConfig the second level is not correctly define.

Fixed version
http://plnkr.co/edit/vleQ7C7Z0VBwrwx46UO3?p=preview

Original version
http://plnkr.co/edit/3ZbRYYzIxiiliP2Gbzm7?p=preview

### DIFF
--- a/src/app/directives/router-active.ts
+++ b/src/app/directives/router-active.ts
@@ -40,13 +40,27 @@ export class RouterActive {
   }
 
   ngOnInit() {
-    this.router.subscribe(() => {
+    this.routerLink.changes.subscribe(() => {
       if (this.routerLink.first) {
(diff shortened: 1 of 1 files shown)


ggaulard:
I forgot to specify the test scenario :s my bad.

Crisis Center ->  Dragon Burning Cities
With Fix Version the "Dragon Burning Cities" Link is yellow 'active' class defined.
With the original version the "Dragon Burning Cities" is not updated.

I think the the ChildRoute doesn't fire event to the subscribe when the RootRouter change.
This may come from a bug in Angular2.

gdi2290:
LGTM

PR #373, opened 2016-03-02
Webworker are broken without parse5

datoml:
Regarding to this [issue](https://github.com/angular/angular/issues/7081) parse5 has to be added to the package.json in an older version to make webworkers work.


gdi2290:
thanks!

PR #429, opened 2016-03-16
Webpack configuration cleanup

meiblorn:
Comments cleanup

It's the same as before

It's the same as before

* **Other information**:
 - Moved some variables to constants. 
 - Reordered plugins initialization headers


gdi2290:
thanks!

PR #431, opened 2016-03-16
remove filename from CommonsChunkPlugin to use filename in output with chunkhash

ktong:

remove filename from CommonsChunkPlugin to use filename in output with chunkhash

build:prod generates ['main', 'vendor', 'polyfills'].bundle.js

build:prod generates ['main', 'vendor', 'polyfills'].[chunkhash].bundle.js

* **Other information**:

It will allow server to cache js files forever.


gdi2290:
thanks!

PR #439, opened 2016-03-18
Escaped file search patterns

Freak613:
Escaped file search patterns in package.json commands
Currently star sign(*) in bash interprets as its special character and finds files only on first nesting level.
Escaped file search patterns so it won't interprets by bash.


gdi2290:
thanks!

PR #447, opened 2016-03-21
start server before running e2e test

ktong:

Feature: start and stop server when running e2e test

To run e2e test, you have to start & stop server by manual.

Automatically start & stop server when running e2e test.

* **Other information**:


gdi2290:
I like it but I rather it be a separate task that also needs to be documented. the tasks also needs to work in windows.

ktong:
Sure. I would like to close it for now

PR #453, opened 2016-03-21
Make Webpack Upgradable and DRY

Spittal:
The webpack configuration has been changed so that components of the configuration that is shared between production and development are now in one file. It makes the files more cohesive and easier to update. 

I've tried introducing this before, but was rejected on grounds of complexity. I would like to argue that this configuration is yes slightly more complex (not anymore complex than other parts of this starter pack) but makes it much easier to to merge in changes from this repo into existing projects. Not only does it allow you put your custom loaders in a single file, but also the "config" object having the ability to be passed around means that you can merge changes from this repo much easier without breaking some "application specific" changes.

There is a lots of duplicate code in the webpack configurations.

Made the webpack conmfiguration more dry, and more scalable.

* **Other information**:
I've been trying to stay up to date with the latest polyfills, and beta versions of Angular while developing an application, It's important to me that I am able to quickly apply changes made in this repo to my own applications. This PR would make that process much faster for me.


gdi2290:
this is fine having a common/base/default config but only if we create a new copy
```es6
var defaultConfig = require('webpack.defaults.js');

module.exports = webpackMerge(defaultConfig, {
  devtool: 'eval'
});
```
that's more familiar to new webpack devs

mbcooper:
+1

Spittal:
Hey @gdi2290 I was able to refactor using webpack merge. Tested with `npm run server` `npm run build:prod` and `npm run build:dev`. Seems to be working great.

fishfillet:
+1

tsm91:
@Spittal @gdi2290 why not `Object.assign`? im not a webpack expert, but the config is just an object

`Object.assign({}, obj1, obj2)` will create a new object

gdi2290:
thanks!

PR #461, opened 2016-03-22
add rxjs dependencies for test

ktong:

Bug fix: test failed when service use map from rxjs

Failed: undefined is not a constructor 

Test passed

* **Other information**:


gdi2290:
good catch

PR #469, opened 2016-03-24
change the order of providers

ktong:

Feature: Register ENV_PROVIDERS after PROVIDERS so that default providers can be override by ENV_PROVIDERS

* **Other information**:


gdi2290:
is there a use case where you need this?

PR #475, opened 2016-03-28
chore(ng2lint): update ng2lint to codelyzer

mgechev:
ng2lint was renamed to codelyzer. This PR reflects the change.


gdi2290:
I like it but why the name change if it's still ng2?

mgechev:
Linting part [sounds optional])https://medium.com/@Jakeherringbone/why-linting-makes-me-yawn-cadbd9a51ca9), in the end we came up with entirely different name. Probably having `ng2` prefix or suffix is a good idea for making the model more recognizable initially but it doesn't sound nice :-).

PR #502, opened 2016-04-12
[Quick Fix] Fixed small typo

KapuzenSohn:
The comments of the webpack test say that html is loaded but css is loaded. The sam is for css where the comment says it loads css but it loads html. This pull Requests fixes this.

Same as before

Same as before

* **Other information**:
In the comment of the file it sayed that the loader load `*.css` but it actually loaded `.html`.
Same is for the `*.html` comment it loads `.css` files


gdi2290:
thanks!

PR #503, opened 2016-04-12
Step-through debugging for tests in Karma

shlomiassaf:
The current step does not allow step through debugging for TS files or any files actually.

The spec-bundle.js file has no .map file so browser support does not allow debugging.

The solution is simple and taken from the karma-webpack README.

https://github.com/webpack/karma-webpack#source-maps

Thanks.


abierbaum:
@shlomiassaf Thanks for the great fix.  This bit me just this morning.

PR #514, opened 2016-04-16
chore: simplify usage of CommonsChunkPlugin

Foxandxss:

It is a chore, simplifies config.

Doesn't change current behavior.

Doesn't change current behavior.

* **Other information**:

You don't need to specify the `main` in here. What webpack is going to do is create a hierarchy (making vendor to be the parent of `main`) and having that `main` there won't add any value. The minChunks is not needed for the current config.

Also, you could drop the reverse helper and just swap the chunks positions.

FWIW, Webpack is smart enough to order the chunks like: polyfills -> vendor -> main so for HtmlWebpackPlugin there is no real need to manually order the chunks.

Cheers.


gdi2290:
LGTM

PR #529, opened 2016-04-21
move configuration of HashLocationStrategy to environment.ts

ktong:

It keeps default setting for LocationStrategy and it's easier to customize for different env.

* **Other information**:

### DIFF
--- a/src/platform/browser/providers.ts
+++ b/src/platform/browser/providers.ts
@@ -3,11 +3,8 @@
  */
 
 
-import {provide} from 'angular2/core';
-
 // Angular 2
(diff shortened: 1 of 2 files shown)


mastertinner:
Thanks for the PR @ktong! Can you please fix it so the tests are green?

ktong:
fixed the lint error. However, still cannot pass the unit tests because test on HEAD is failed too.

ktong:
redo it base on angular beta16

ktong:
the test will pass after apply https://github.com/AngularClass/angular2-webpack-starter/pull/551

PR #537, opened 2016-04-24
Upgrade copy-webpack-plugin to 2.1.3

cortopy:
When upgrading copy-webpack-plugin, an error stops webpack-dev-server if devServer.outputPath is not given


mastertinner:
thanks, @cortopy!

PR #550, opened 2016-04-28
Optimize Images and Minify Html for Prod

t0rrro:
add html-minify-loader, image-webpack-loader to further optimize production builds.


t0rrro:
Closing for better html minifier

PR #581, opened 2016-05-11
simplify HtmlWebpackPlugin options

cebor:
`chunksSortMode: 'dependency'` should sort your injects correctly, without helper function, also will be the default behavior in webpack2


gdi2290:
thanks!

PR #613, opened 2016-05-19
Material remove

ebeal:
Remove material2 and all references from starter app

* **Other information**:
Also update tests and simplify css


gdi2290:
LGTM

IAMtheIAM:
Why did you do this? This should be reverted because it just strips down the template without benefit. I think it was cool to show off the new material design. It's easy to remove material if we don't want it, but harder to add if we do.

JulioC:
I like this move, makes the example app much more concise.

I just want to understand the reasoning behind this, since even today we had commits in the direction of improving material support. Should I expect this PR to be reverted anytime soon?

zigzag95:
Material is a part of angular for now. I don't understand this PR

alvipeo:
I actually agree with the remove. If you try to actually build a real app you can't with material. It's way too early to use it.

dsebastien:
Ideally what we should move towards is:
* bare project with the basic necessities
* yeoman generators (basic, options to add material, etc)

Ideally (that's what we're doing at work):
* externalize all the build to a separate project (i.e., make it reusable but still extensible)
* externalize all build scripts to a CLI

But that takes some effort idd :)

jaycangel:
Hi, can I ask why material was removed? One of the reasons i wanted to use this starter pack is because it had angular 2 and material. Is there another way you suggest I incorporate material back in?

What other css framework should we be using if not material?

abdulhaq-e:
mgechev's seed has a an "[Examples](https://github.com/mgechev/angular2-seed#examples)" section where it shows forked repos that added library X to the seed

Perhaps a similar thing is needed for this seed because no seed will suite everyone!

JulioC:
I believe that's the plan with the [material2 branch](https://github.com/AngularClass/angular2-webpack-starter/tree/material2). If kept in sync with master, that should be enough.

IAMtheIAM:
Yes, please update [material2 branch](https://github.com/AngularClass/angular2-webpack-starter/tree/material2) and this will be a fine solution to meet more dev's needs.

@jaycangel You can check out the material2 branch when its ready, or simply go to master branch,  and look at previous versions of [package.json and src/platform/browser/angular2-material2/index.ts](https://github.com/AngularClass/angular2-webpack-starter/commit/ddede6dbb65ca32af1db743429294cbf519aab23)

ebeal:
Hey guys sorry for the lack of explanation about this. Yes the [material2 branch](https://github.com/AngularClass/angular2-webpack-starter/tree/material2) will be kept in sync with master. The idea is to keep material separate especially while it is still unstable (while still making improvements and expanding on examples on that branch). @gdi2290 will make an update to the readme about this so people can still use that version if they like.

shlomiassaf:
👍 

Good move, simplicity in mind.

PR #634, opened 2016-05-29
Style 03-05: custom TSLint rule to enforce style

dweitz43:
feature

Angular Style Guide Rule 03-05 Import Destructuring Spacing must currently be enforced manually by the user.

a custom TSLint rule has been implemented to help attempt to enforce this style guide rule as part of the linting process.

* **Other information**:
I didn't have any prior experience with TSLint before this, so I used these as resources to help create this custom rule:
-[TSLint custom rules](http://palantir.github.io/tslint/develop/custom-rules/)
-the visitImportDeclaraction function in the TSLint [whitespaceRule.ts](https://github.com/palantir/tslint/blob/master/src/rules/whitespaceRule.ts)
-[codelyzer](https://github.com/mgechev/codelyzer) was also very helpful and i'm not sure if this might be more suitable to include in that project as well. it doesn't feel like the creation of a custom tslint rule to enforce style is necessarily inside the scope of this starter, but this was the best way i think i could come up with to try to enforce this style as part of the tslint process.

### DIFF
--- a/config/rules/importDestructuringSpacingRule.js
+++ b/config/rules/importDestructuringSpacingRule.js
@@ -1,7 +1,7 @@
 /**
  * custom TSLint rule: Import Destructuring Spacing Rule
  * rules written in TypeScript must be compiled to JavaScript before invoking TSLint. to compile:
- * tsc -m commonjs --noImplicitAny importDestructuringSpacingRule.ts ../node_modules/tslint/lib/tslint.d.ts
+ * tsc -m commonjs --noImplicitAny importDestructuringSpacingRule.ts ../../node_modules/tslint/lib/tslint.d.ts
  *
  * enforces Angular Style Guide Rule 03-05: Import Destructuring Spacing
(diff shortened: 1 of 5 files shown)


gdi2290:
thanks!

Foxandxss:
I am totally going to steal this for codelyzer. So we can all get use of that.

Thanks for creating it.

dweitz43:
@Foxandxss I was thinking about trying to add it there as well, but I have been a little busy at my job lately. 

Let me know if you need/want any help with that.

JonnyBGod:
I am getting lint errors due to this when doing ```import * as D3 from 'd3';```.
We probably need to add some test to skip this import format from this rule.

Foxandxss:
Uhh, that is a good catch mate. I am totally creating an issue at codelyzer.

Foxandxss:
Ok, it is fixed on latest codelyzer, we just need to bump it here.

dweitz43:
thanks guys, not sure how i neglected that

PR #636, opened 2016-05-29
update dependencies

mastertinner:

* **Other information**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
 
     "protractor": "^3.2.2",
 
-    "parse5": "^1.5.1",
+    "parse5": "^2.1.5",
     "phantomjs-polyfill": "0.0.2",
     "phantomjs-prebuilt": "^2.1.7",
     "remap-istanbul": "^0.6.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -101,7 +101,7 @@
 
     "protractor": "^3.2.2",
 
-    "parse5": "^1.5.1",
+    "parse5": "^2.1.5",
```
gdi2290: we can't update this

mastertinner: why is that? i have been working with 2.1.5 in all my projects without any problems and without getting any errors/warnings about that

gdi2290: someone added this for universal support which only supports 1.x.x it really shouldn't be in the project tbh because it's only needed if someone installs universal

gdi2290: can you also add me to your forked repo that way I edit your pull-requests?

mastertinner: Then let's remove it^^ Unfortunately I destroy the fork after every PR you merge and fork again for every PR.

PR #645, opened 2016-06-01
GitHub deployment script into gh-pages - dev or prod builds.

shlomiassaf:
feature

Same as ng-cli, an option to deploy to github into the `gh-pages` branch.
This feature requires some modifications to `index.html` and an additional webpack config file `webpack.github-deploy.js`. 

* **Other information**:
Dependes on #643 
This PR adds the `gh-pages` npm packages to the devDependencies.
`gh-pages` has a lot of option to customise the operation, for this initial PR I chose to allow a limited set of configuration not to confuse.
Use can change the git remote origin and a constant commit message, see `webpack.github-deplot.js`

### DIFF
--- a/config/github-deploy/index.js
+++ b/config/github-deploy/index.js
@@ -1,0 +1,59 @@
+const helpers = require('../helpers');
+const execSync = require('child_process').execSync;
+
+const REPO_NAME_RE = /Push  URL: https:\/\/github\.com\/.*\/(.*)\.git/;
+
+function getWebpackConfigModule() {
+  if (helpers.hasProcessFlag('github-dev')) {
+    return require('../webpack.dev.js');
+  } else if (helpers.hasProcessFlag('github-prod')) {
+    return require('../webpack.prod.js');
+  } else {
+    throw new Error('Invalid compile option.');
+  }
+}
+
+function getRepoName(remoteName) {
+  remoteName = remoteName || 'origin';
+
+  var stdout = execSync('git remote show ' + remoteName),
+      match = REPO_NAME_RE.exec(stdout);
+
+  if (!match) {
+    throw new Error('Could not find a repository on remote ' + remoteName);
+  } else {
+    return match[1];
+  }
+}
+
+function stripTrailing(str, char) {
+
+  if (str[0] === char) {
+    str = str.substr(1);
+  }
+
+  if(str.substr(-1) === char) {
+    str = str.substr(0, str.length - 1);
+  }
+
+  return str;
+}
+
+/**
+ * Given a string remove trailing slashes and adds 1 slash at the end of the string.
+ *
+ * Example:
+ * safeUrl('/value/')
+ * // 'value/'
+ *
+ * @param url
+ * @returns {string}
+ */
+function safeUrl(url) {
+  const stripped = stripTrailing(url || '', '/');
+  return stripped ? stripped + '/' : ''
+}
+
+exports.getWebpackConfigModule = getWebpackConfigModule;
+exports.getRepoName = getRepoName;
+exports.safeUrl = safeUrl;
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -59,7 +59,7 @@ module.exports = {
     'main':      './src/main.browser.ts'
 
   },
-
+  
   /*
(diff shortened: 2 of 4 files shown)


gdi2290:
+1000

gdi2290:
let's wait a bit for feedback

shlomiassaf:
The main focus for this discussion is on the index.html file

In github pages there is a prefix added to every project, which is the repository name.
This requires modification to the `index.html` so every `href` and `src` links are properly set.

The current PR isn't ideal since it adds noise to the index.html file.

I'm not `<base>` expert but I think it might be done with only changing the `<base>`, this should effect all paths automatically. I have tried to played with it but couldn't get consistent result and I guess it's because I was missing a slash or something.

The question is, does `<base>` works the same in all browsers? In chrome I got it to work on the javascript files but the static `<link>` tags did not get the base prefix.

Also, does the location of the `<base>` tag matters? I mean, if it's after a `<link>` tag it won't operate on it?

shlomiassaf:
If anyone want to see a working sample of this PR:

**Site:**
http://shlomiassaf.github.io/ng2-chess/

**Dist output:**
https://github.com/shlomiassaf/ng2-chess/tree/gh-pages

**Index.html**
https://github.com/shlomiassaf/ng2-chess/blob/gh-pages/index.html

valorkin:
hey, looks really good
but do you really need to do additional config file or just use
https://www.npmjs.com/package/gh-pages

deploys as simple as:
`gh-pages -d dist`

shlomiassaf:
@valorkin The implementation does use the `gh-pages` package but it wont work out of the box, it need some help getting it to actually work.

The main reason is that in github the base url is always **username**.github.io/**repo-name**. In most cases the application will not have a **repo-name** prefix, it's usually the root of the domain.

So, we have 2 different setups, one for a deployment in production and one for deployment in github.
This is why we need an additional config since it a different setup.

We also need to think about the purpose of `angular2-webpack-starter`, it's a set tools aimed to making the developers life easier, not having to manually manage work flows.
Having to run `gh-pages -d dist` manually requires some previous steps, you have to run the build and if you forgot you probably just posted an older version to github, this is a common mistake that happened to me more then once :)
It get's even worse if you forgot to edit your 'index.html' to support the github pages base url prefix, again happened to me more then once.
This is actually why I created this flow, then realized it can be integrated into this project.

Another thing is that you can deploy 2 (or more, custom) compilation artefacts to github, minified production build or a development build.

Also, this PR tries to keep it simple so you don't have a lot of configuration options to control the `gh-pages` package, this can be done in a later phase. I prefer a simple PR first so we can understand it.

So using the `gh-pages -d dist` approach you need to do some manual work, which is the same for every release: build and edit `index.html`, this is a prefect scenario for automation.

If you investigate the config file you will see it's basically a simple webpack plugin, I think we should first adopt the PR and then opt for a plugin after we get around most use cases and options to support.

shlomiassaf:
The main issue here is the noise in **index.html**

Having to prefix every resource:
```
<link rel="apple-touch-icon" sizes="57x57" href="<%= webpackConfig.output.publicPath %>assets/icon/apple-icon-57x57.png">
```

Add's a lot of noise that I really don't like, and it's "sort of" an API change, this is why I asked @gdi2290 to ask for community help.

gdi2290:
I merged the other PR. do you think you can rebase? I like the static asset array if you can build that out I'll merge it

shlomiassaf:
@gdi2290 before I start, some consult.

Since a static asset file has volume I don't think it should be within the configurations themselves.

We can put it in a file, in the config directory... something like
`assets.common.json` or `assets.dev.json` etc... 

In the configs we can opt in for them if the file exists and create the html, this result's in a non breaking API modification.

Another option is to let the user set the json asset file in the config's.

I like the first option better, it's convention over configuration which is less confusion.

Comments please...

shlomiassaf:
@gdi2290 Once #679 is merged I'll rebase this, works like a charm.

shlomiassaf:
@gdi2290 please have a quick final look.

Github pages output can be found here:
http://shlomiassaf.github.io/angular2-webpack-starter/
https://github.com/shlomiassaf/angular2-webpack-starter/tree/gh-pages

PR #647, opened 2016-06-01
docs(readme/quick-start): remove "global install" section, it's not r…

ajoslin:
Webpack, rimraf, and webpack-dev-server are all already included in package.json as devDependencies.

This means they'll be in node_modules/.bin.

npm scripts implicitly have node_modules/.bin in their path, so the npm scripts will then read from node_modules/.bin, making the global install extraneous.


gdi2290:
while true. without this I'll get 20 issues from windows users

PR #666, opened 2016-06-07
chore: use import spacing rule from codelyzer

Foxandxss:
This rule is now part of codelyzer, so I activated codelyzer again on this project and used the rule. That way if we update it, you don't need to update it here too.


gdi2290:
thanks!

Issue #673, opened 2016-06-09
AppState always return full state object if property is a boolean with value "false"


leonardlin:
- **I'm submitting a ...**
  [X ] bug report
  [ ] feature request
  [ ] question about the decisions made in the repository
- **What is the current behavior?**
  Given this constructor

```
constructor(
    public appState: AppState) {
  }
```

execute the following code:

```
this.appState.set('loggedin', false);
console.log(this.appState.get('loggedin'));

```

It returns: object { loggedin: false }
- **What is the expected behavior?**

it should return just a boolean : false
- **What is the motivation / use case for changing the behavior?**

Fix is to change the getter function this way

```
get(prop?: any) {
    // use our state getter for the clone
    const state = this.state;
    return state.hasOwnProperty(prop) ? state[prop] : state;
  }

```
- **Please tell us about your environment:**
- Angular version: 2.0.0-beta.X
- Browser: [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
- **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

leonardlin:
I can create a PR if that helps, let me know

gdi2290:
can you make a PR, thanks!

leonardlin:
created PR: https://github.com/AngularClass/angular2-webpack-starter/pull/675

boldtrn:
Should be closed, since PR was merged?

PR #675, opened 2016-06-10
bool false value not returned, Resolves issue #673

leonardlin:
resolves issue #673 

if the property is of type boolean and value false then the whole state object is returned

the property value is returned, even if the value is false

* **Other information**:

this resolves the issue #673, that boolean properties with value 'false' are not returned.


gdi2290:
thanks!

PR #747, opened 2016-07-03
docs(readme): fix typings install command

acdcjunior:

Simple readme fix, to a command that works.

`npm run typings-install` yields:

> npm ERR! missing script: typings-install

Typings will be installed correctly.

* **Other information**:

...since there's no `typings-install` script @ package.json.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ npm install
 
 # WINDOWS ONLY
 # install typings
-npm run typings-install
+npm run typings install
 
 # start the server
 npm start


gdi2290:
it would have to be npm run typings -- install

acdcjunior:
Uhm, strange. `npm run typings install` does work for me.

`npm run typings -- install` works as well (and may be more idiomatic, I don't know, really). I was just under the impression that the ` -- ` was only required if the arguments had a `--` prefix (e.g. `--install` instead of just `install`).

Anyway, I see you fixed the `package.json`. All is good. Thanks as always. Cheers!

Issue #863, opened 2016-08-17
Automatic splash-screen/icon generation from template


stevemartin:
**Note: for support questions, please use one of these channels:** [Chat: AngularClass.slack](http://angularclass.com/member-join/) or [Twitter: @AngularClass](https://twitter.com/AngularClass)

* **I'm submitting a ...**
[ ] bug report
[x ] feature request
[ ] question about the decisions made in the repository

* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
I didn't see any feature in the webpack build ( but could be wrong of course ) for auto generating splash screens and icons for different devices from a template `png` or similar.

* **What is the motivation / use case for changing the behavior?**
I would like to be able to generate device specific splash screens and icons from a single template file when I `npm start` or `npm build:xxx` the project.

* **Please tell us about your environment:**

- Angular version: 2.0.0-beta.4
- Browser: [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

#864

PR #864, opened 2016-08-17
#863 Automatic icon generation from template file

stevemartin:
Generates icon files from a single icon template when you build

It doesn't do this

Generates icon files from a single icon template when you build

* **Other information**:
#863

* Uses `splashicon-generator` npm package
* Create plugin for webpack
* Edit `src/assets/icon/icon-template.png` to change the icon


katallaxie:
👍 Can you move the plugin to an `npm` package? And can we then reference it in the `README`. There are some projects which have somehow different icons for different sizes.

Issue #898, opened 2016-08-25
Error in latest checkin which used webpack2


StefH:
* **I'm submitting a**
[ x] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Solution**
The project.json is missing:
`"string-replace-loader": "^1.0.3"`

Luchillo:
Made a pull request: https://github.com/AngularClass/angular2-webpack-starter/pull/900

StefH:
Solved by https://github.com/AngularClass/angular2-webpack-starter/pull/899

PR #899, opened 2016-08-25
Include string-replace-loader as devDependency

isaacplmann:

Adds string-replace-loader as a devDependency in package.json

Errors are thrown when building - since string-replace-loader is missing.

Fixes the errors.

* **Other information**:

I like cheese.  :-)


isaacplmann:
Fixes #898

Issue #728, opened 2016-06-24
[Feat] Static template compilation, Web Workers, Service Workers and sw-precache


luchillo17:
- **I'm submitting a ...**
  [x] feature request
- **Do you want to request a _feature_ or report a _bug_?**
  A feature
- **What is the current behavior?**
  There's no current support for [Static Template Compilation](http://teropa.info/blog/2016/05/19/things-that-excite-me-about-angular-2.html#static-template-compilation) nor [Web Workers](http://teropa.info/blog/2016/05/19/things-that-excite-me-about-angular-2.html#web-worker-support) nor [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API)
- **What is the expected behavior?**
  Have support for those 3 technologies in the app.
- **What is the motivation / use case for changing the behavior?**
  - Static template compilation: Speed up the render process and performance as well as lowering the size of the bundles.
  - Web Workers: Offloading resource intensive operations to a new thread thus making better use of multicore cpus (today even mobile have multicore).
  - Service workers: Mainly to support offline applications using libs like [sw-precache](https://github.com/GoogleChrome/sw-precache) or [sw-toolbox](https://github.com/GoogleChrome/sw-toolbox), but also to support background sync and push notifications.  
- **Other information**
  Here are some links about those subjects:
  - [Static Template Compilation](http://teropa.info/blog/2016/05/19/things-that-excite-me-about-angular-2.html#static-template-compilation)
  - [Web Workers](http://teropa.info/blog/2016/05/19/things-that-excite-me-about-angular-2.html#web-worker-support)
  - [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API)
  - [The difference between Service Workers, Web Workers and WebSockets](https://aarontgrogg.com/blog/2015/07/20/the-difference-between-service-workers-web-workers-and-websockets)

Aditionally i'm working on the sw and sw-precache feature myself so i can provide some help in that field.

Luchillo:
I've been digging in how to integrate with webpack and so far i have seen 2 plugins that takes care of that:
- [Offline plugin](https://github.com/NekR/offline-plugin)
- [sw-precache-webpack-plugin](https://github.com/goldhand/sw-precache-webpack-plugin)

So far it seems the Offline plugin is better since it states that it fallbacks to AppCache if service worker isn't supported in the browser, further testing and analysis required.

Luchillo:
[...]

PR #902, opened 2016-08-25
Add support for Offline mode with NekR/offline-plugin

Luchillo:
This PR will add support for offline mode with this plugin: [NekR/offline-plugin](https://github.com/NekR/offline-plugin).

Relate to https://github.com/AngularClass/angular2-webpack-starter/issues/728 for the discussion about which plugin to choose as well as other very desirable features.


katallaxie:
Good work. I would be great to have this as patch to the start. So in case one would want to have this feature, without making it the standard.

Luchillo:
2 questions, i do think this is a nice default, but as you say it isn't for everyone, so how do we make it a patch?

What's that about checks? is that the testing of this repo with travis? if so i think it was already broken before i started to edit this feature.

NekR:
You probably don't need `offline-plugin` in dev-mode, only in production. It's especially useless with `wepback-dev-server`. Production build anyway must be tested locally too, so SW/AppCache could be tested at that moment.

Luchillo:
Sooo, using the plugin runtime only in prod mode, and adding tests? to tell you the truth i haven't mess around a lot with testing in JS, i have done it in Ruby and Rails but not in JS, so i don't feel like making tests for that, i'll probably make very bad tests and you guys will end up rewriting them.

NekR:
What's happening with this issue anyway?

Luchillo:
Oh i'm sorry about that story feed up there, i'm just trying to update to V6 branch with the Offline plugin working, however i'm having issues with the DLL's to be included in the browser cache.

katallaxie:
@Luchillo I close the PR, because `v6` is not yet destined to move to `master`. You should push to my `v6` if it works.

Luchillo:
Actually i didn't want to push all that, i tough the only thing that get's in the PR is when i make the PR, not later commits, so all along it was commits not intended for this PR.

katallaxie:
Everything that is in the branch is this branch and not in the master is merged.

PR #925, opened 2016-08-29
add material component to Detail Module

dweitz43:
branch update

Detail module does not include material components

implement material component in Detail Module, update to master branch as well

* **Other information**:


src/app/+detail/index.ts
```diff
@@ -20,8 +20,9 @@ export const routes = [
     BrowserModule,
     FormsModule,
     RouterModule.forChild(routes),
+    MdModule.forRoot()
```
zigzag95: You don't need to use .forRoot() on MdModule since md providers are already loaded by app

tsconfig.json
```diff
@@ -17,7 +17,6 @@
       "es6"
     ],
     "types": [
-      "core-js",
```
zigzag95: I don't get any conflict and removing this I get errors Promise Unknown

src/app/+detail/index.ts
dweitz43: @zigzag95 i wasnt sure about that for a lazily loaded module, thanks for the tip! updating the pr now...

tsconfig.json
dweitz43: @zigzag95 i think this was a result of the merge with master, as i believe it is not in tsconfig.json in the master branch....however i didnt get any errors removing it, but i am adding it back now

dweitz43:
@gdi2290 seems like the build fails because of a problem installing node-sass, but i dont experience this issue at all... i wonder if this can still be merged? or if i should just wait until the repo is upgraded to rc.6 now that it has been released?

gdi2290:
@katallaxie I avoid any modules that might cause problems when installing for example node-sass and CompressionPlugin

PR #928, opened 2016-08-30
Add missing type information to compile in case of `noImplicitAny`

kondi:

Added missing type information to compile in case of `noImplicitAny` is set to true

If you set `noImplicitAny` to true, you have many TypeScript errors.

Compiled without errors when `noImplicitAny` is set to true.

* **Other information**:

See https://github.com/angular/angular-cli/issues/107#issuecomment-164809092

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -128,9 +128,9 @@ module.exports = {
       {
         test: /\.ts$/,
         loaders: [
+          '@angularclass/hmr-loader',
           'awesome-typescript-loader',
-          'angular2-template-loader',
-          '@angularclass/hmr-loader'
+          'angular2-template-loader'
         ],
         exclude: [/\.(spec|e2e)\.ts$/]
       },
--- a/src/custom-typings.d.ts
+++ b/src/custom-typings.d.ts
@@ -56,8 +56,8 @@ interface SystemJS {
 }
 
 interface GlobalEnvironment {
-  ENV;
-  HMR;
+  ENV: string;
+  HMR: boolean;
   SystemJS: SystemJS;
   System: SystemJS;
 }
--- a/src/app/about/about.component.ts
+++ b/src/app/about/about.component.ts
@@ -27,7 +27,7 @@ console.log('`About` component loaded asynchronously');
   `
 })
 export class About {
-  localState;
+  localState: any;
   constructor(public route: ActivatedRoute) {
 
   }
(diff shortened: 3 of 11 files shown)


src/custom-typings.d.ts
```diff
@@ -43,9 +43,6 @@ import * as _ from 'lodash'
  *
  */
 
-// support NodeJS modules without type definitions
-declare module '*';
```
gdi2290: we need this for typescript 2.0 so the experience for installing new modules isn't so crazy.

kondi: For some reason I have got compiler error with that line. I have checked now and not true anymore...
I have just removed this change from the pull request (so now the wildcar declaration is there).

config/webpack.common.js
```diff
@@ -128,9 +128,9 @@ module.exports = {
       {
         test: /\.ts$/,
         loaders: [
+          '@angularclass/hmr-loader',
           'awesome-typescript-loader',
-          'angular2-template-loader',
-          '@angularclass/hmr-loader'
+          'angular2-template-loader'
```
gdi2290: why was the order changed?

src/custom-typings.d.ts
gdi2290: this is fine. I rather have developers who are new to typescript not run into problems when installing 3rd party modules. you can add a comment in the readme about removing this line if they use `noImplicitAny`

config/webpack.common.js
kondi: It was very confusing for me when the TypeScript compiler started to report errors in line/column position which did not even exist in the source file. It took some time for me to realize the angular2-hmr-loader injected new lines into the code.

The injected code in angular2-hmr-loader is not typed and does not use any TypeScript extension, it is plain JavaScript. Because of that, it would make more sense to apply it to the JavaScript code (output of awesome-typescript-loader) and not the TypeScript source. Why to push is through the type checker when we know it is not typed?

If I keep the original order, the angular2-hmr-loader is applied before the TypeScript compilation. In case of noImplicitAny you will have type errors in the injected code, but you do not have any control of them.

Another option is to annotate the injected code in angular2-hmr-loader with type information, but in this case it would not be anymore JavaScript compatible. However if you think that is not a problem and the target of angular2-hmr-loader is always TypeScript source, then I can make a pull request in that project to add the missing types (probably few basic any is enough though).

gdi2290: thanks, I'll update angular2-hmr-loader to also emit ts code later if devs want it to transform ts code rather than es5

PR #944, opened 2016-09-01
Updating to angular2-*@2.0.0-rc6 (deps, specs ...)

katallaxie:

Angular 2 RC6

Angular 2 RC5

* **Other information**:


katallaxie:
@gdi2290 I have updated to RC6. There seems to be an issue with the `DedupePlugin`. Should we move ahead anyway?

gdi2290:
@katallaxie there always seems to be a problem with `DedupePlugin`. I would leave a comment saying it doesn't work atm otherwise developers would be confused if they uncomment it

src/app/+detail/index.ts
```diff
@@ -17,7 +17,6 @@ export const routes = [
     Detail
   ],
   imports: [
-    BrowserModule,
     FormsModule,
```
gdi2290: don't you need `CommonModule` here?

katallaxie: I have included it again. But `BrowserModule` is already imported in `app.module.ts`.

PR #952, opened 2016-09-03
Add Webpack DLL

zigzag95:
>npm run dll
>nom run server:dll


zigzag95:
I need to manage both index.html to not break current behaviours 
I'll make a new PR

Issue #975, opened 2016-09-09
include HardSourceWebpackPlugin


gdi2290:
https://github.com/mzgoddard/hard-source-webpack-plugin
cc @zigzag95

PR #981, opened 2016-09-11
feat: Add HardSourceWebpackPlugin

zigzag95:
Add HardSourceWebpackPlugin in dev webpack config

Reduce build time on src files change. Only impacted modules are rebuilt.

* **Other information**:
cf https://github.com/mzgoddard/hard-source-webpack-plugin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,7 @@ npm-debug.log
 # IDE #
 .idea/
 *.swp
+
+# hard-source-cache #
+hard-source-cache/
+records.json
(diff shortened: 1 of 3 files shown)


zigzag95:
@gdi2290 need comments

katallaxie:
@zigzag95 seems to be a case for the docs

d3viant0ne:
I agree with @katallaxie, this is something that should be documented and not added.

MarkPieszak:
@d3viant0ne Pat actually wanted this one in the Starter, see here: https://github.com/AngularClass/angular2-webpack-starter/issues/975

gdi2290:
I see no downside to this unless someone can think of one. If this introduces gotchas for beginners then I would vote for moving it to docs

Issue #986, opened 2016-09-12
wrong module config key for webpack2


akomm:
[this](https://github.com/AngularClass/angular2-webpack-starter/blob/2f614b45851056c2ec463133267cb7f8c36c994d/config/webpack.common.js#L85) is wrong. 

It has changed in webpack2: [link](https://github.com/webpack/webpack/blob/dc50c0360e87204ea77172910e877f8c510f3bfb/lib/WebpackOptionsDefaulter.js#L74)

key webpack1: 'resolve.modulesDirectories' key in webpack2: 'resolve.modules'

PR #987, opened 2016-09-12
fix/986

akomm:
fix for #986


katallaxie:
In Webpack 2, resolvers from `root`, `modulesDirectories`, and `fallback` settings merge into a single property – `modules`. You should add these to the `modules`.

akomm:
I'v noticed the root option missing too, but could not figure out via history when that happened. There seem to be no `fallback` option in use by default, so only root is needed. Also I think root does not need to be absolute anymore (the helper), since `modules` does not require it.

PR #1005, opened 2016-09-14
Set karma-webpack to only log errors

ashleahhill:

Not sure when the api changed, but `webpackServer: { noInfo: true },` doesn't work anymore. This uses 
`webpackMiddleware: { stats: 'errors-only'}` to accomplish the same goal.

When running tests, built modules are output to the console.

Karma-webpack will only output errors during code compile for tests.

* **Other information**:

Set wepack to only output errors during testing


katallaxie:
LGTM

PR #1018, opened 2016-09-19
Pins webpack version to 2.1.0-beta.21

dangerdespain:
Prevents breakage from the latest webpack beta versions (2.1.0-beta.22 and up)

See below console output below

Pins the webpack version to the last working version until a webpack ninja can upgrade the webpack config files.

* **Other information**:

Here's my console output using webpack 2.1.0-beta.22

`> webpack-dev-server --config config/webpack.dev.js --progress --profile --watch --content-base src/

/Users/me/clients/delve/dash/node_modules/webpack/lib/webpack.js:16
		throw new WebpackOptionsValidationError(webpackOptionsValidationErrors);
        ^
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration has an unknown property 'tslint'. These properties are valid:
   object { amd?, bail?, cache?, context?, devServer?, devtool?, entry, externals?, loader?, module?, name?, dependencies?, node?, output?, plugins?, profile?, recordsInputPath?, recordsOutputPath?, recordsPath?, resolve?, resolveLoader?, stats?, target?, watch?, watchOptions? }
 - configuration.module has an unknown property 'preLoaders'. These properties are valid:
   object { rules?, loaders?, noParse?, unknownContextRequest?, unknownContextRegExp?, unknownContextRecursive?, unknownContextCritical?, exprContextRequest?, exprContextRegExp?, exprContextRecursive?, exprContextCritical?, wrappedContextRegExp?, wrappedContextRecursive?, wrappedContextCritical? }
   Options affecting the normal modules (`NormalModuleFactory`).
`


yuanfeiz:
Thanks, BTW why did webpack introduce backward incompatible update in the beta version?

SonlyF5020:
2.1.0-beta.22 works for me , 2.1.0-beta.22 is broken

d3viant0ne:
beta.24 has already been tagged which includes a fix for the validations issue.

likerRr:
I think we shouldn't pin older version, instead we should fix fit new webpack's requirements. Here are the examples:

1. [beta.22 => beta.23 breaking changes](https://github.com/angular/angular-cli/pull/2237/files#diff-427f56af65fe1d070dcca914a7f884d6L19)
2. [Other updates](https://github.com/webpack/webpack.js.org/blob/develop/content/how-to/upgrade-from-webpack-1.md#moduleloaders-is-now-modulerules)

I tried to update configurations locally, but since the [`metadata`](https://github.com/AngularClass/angular2-webpack-starter/blob/master/config/webpack.common.js#L43) is now a custom option (which webpack >= beta.23 doesn't support) I don't know where to put it. There are no problems with other options.

tobyforever:
22 works for me. 23& 24 failed.

charlieschen:
2.1.0-beta.22 link is missed, now if you set 2.1.0-beta.21 or 2.1.0-beta.22 both direct up upgraded tp 2.1.0-beta.24, but start version hadn't 2.1.0-beta.24 new config for webpack.common.js

PR #1049, opened 2016-09-24
Remove @types/corejs

harunurhan:

Cleanup

`@types/corejs` is installed but not used due to recent changes.

Less dependencies, cleaner `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,6 @@
     "zone.js": "~0.6.17"
   },
   "devDependencies": {
-    "@types/core-js": "^0.9.34",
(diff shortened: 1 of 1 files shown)


zigzag95:
Are you sure?

xielongjiang:
@harunurhan you remove @types/corejs,will lead to WebStrom display error.because i use webstrom

harunurhan:
@xielongjiang Since `corejs` provided by `typescript` itself by setting `lib` in `compileOptions` of `tsconfig.json`, this shouldn't be necessary and it is the bug should be fixed by WebStorm.

PR #1055, opened 2016-09-26
development oriented dockerfile + docker-compose for easy usage

benyitzhaki:
a new dockerfile for development, running nodejs. mounted the src folder to the local folder for easy development

currently there is just a dockerfile for production usage. as default, the url will be 0.0.0.0:3000

allows developing using dockerfile and docker-compose

* **Other information**

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -1,0 +1,15 @@
+# This image will be based on the official nodejs docker image
+FROM node:latest
+
+# Set in what directory commands will run
+WORKDIR /home/app
+
+# Put all our code inside that directory that lives in the container
+ADD ./ /home/app
+
+# Install dependencies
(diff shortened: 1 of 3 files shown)


d3viant0ne:
Agree with @katallaxie - While I docker everything under the sun, many still use vagrant or don't containerize at all. This as examples in the readme would accomplish the same thing while allowing flexibility to implement without the merge headaches as everyone's compose stack is going to vary significantly.

d3viant0ne:
@benyitzhaki - This type of information is valuable & should be in a readme / docker.md. Given the situational nature of Docker, it's not something I would merge into a starter repo.

We would on the other hand certainly appreciate a doc covering your proposed Docker Stack configuration as a starting point for those that are new to Docker

Issue #1070, opened 2016-09-29
robots.txt and humans.txt are in wrong folder


MZellhofer:
**Note: for support questions, please use one of these channels:** [Chat: AngularClass.slack](http://angularclass.com/member-join/) or [Twitter: @AngularClass](https://twitter.com/AngularClass)

* **I'm submitting a ...**
[X] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Do you want to request a *feature* or report a *bug*?**
bug

* **What is the current behavior?**
`robot.txt` and `humans.txt` are inside `assets` folder

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
If you run npm build, robots.txt and humans.txt are placed inside the assets folder.

* **What is the expected behavior?**
If I am not completely wrong, it would be better to place them in the root folder of the distribution, `dist`.

PR #1076, opened 2016-09-30
Copy humans.txt and robots.txt to dist/

nchutchind:
Bug fix

#1070 humans.txt and robots.txt get copied from the assets directory to the output assets directory.

All files in the assets folder get copied to dist/assets except for humans.txt and robots.txt.
humans.txt and robots.txt get copied from the assets folder to the root dist/ folder.

* **Other information**:

Fixes issue #1070

The standard for humans.txt and robots.txt says that they should be available from the root of the application. This change will make sure that they get copied from the assets folder to the dist/ root instead of to dist/assets.

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -244,6 +244,16 @@ module.exports = function(options) {
       new CopyWebpackPlugin([{
         from: 'src/assets',
         to: 'assets'
+      }], {
+        ignore: [
+          'humans.txt',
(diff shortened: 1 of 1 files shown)


nchutchind:
In the future, if we need to add any more files that get copied to root, it would probably be a lot easier to make a new directory in the assets directory for all of them and just copy the contents instead of individual files.

PR #1089, opened 2016-10-06
fix: github-deploy fix using webpack 2

### LINKED ISSUE #1078: Github-deploy stopped working with webpack 2.x
- **I'm submitting a ...**
  [x] bug report
  [ ] feature request
- **Do you want to request a _feature_ or report a _bug_?**

bug
- **What is the current behavior?**

```
npm run github-deploy
```

throws an error:

```
Unknown argument: github-dev
```
- **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
  https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:AvJOMERrnz94ekVua0u5).

```
npm run github-deploy
```
- **What is the expected behavior?**

Expected to deploy the app to github
- **Please tell us about your environment:**
- Webpack 2.x
- **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

https://github.com/webpack/webpack/issues/2254

shlomiassaf:
Webpack 2 does not accept arbitrary cli arguments, this is why github script fails.

This PR moves to a file based env selection.

closes #1078


package.json
```diff
@@ -24,8 +24,8 @@
     "docs": "npm run typedoc -- --options typedoc.json --exclude '**/*.spec.ts' ./src/",
     "e2e:live": "npm run e2e -- --elementExplorer",
     "e2e": "npm run protractor",
-    "github-deploy:dev": "webpack --config config/webpack.github-deploy.js --progress --profile --github-dev",
-    "github-deploy:prod": "webpack --config config/webpack.github-deploy.js --progress --profile --github-prod",
+    "github-deploy:dev": "webpack --config config/webpack.github-deploy.js --progress --profile",
+    "github-deploy:prod": "webpack --config config/webpack.github-deploy.prod.js --progress --profile",
```
katallaxie: Could you use an environment variable or a `process.argv`, or else. Would be nice to have everything in one config.

Andrey-Pavlov:
I'm using bootstrap-loader. For build:prod works but github-deploy
```
ERROR in ./~/bootstrap-loader/lib/bootstrap.loader.js!./~/bootstrap-loader/no-op.js
Module build failed: TypeError: sassLoader.split is not a function
    at exports.default (D:\FunnyMapNotes\node_modules\bootstrap-loader\lib\utils\processStyleLoaders.js:37:38)
    at Object.module.exports.pitch (D:\FunnyMapNotes\node_modules\bootstrap-loader\lib\bootstrap.loader.js:177:91)
 @ ./~/bootstrap-loader/loader.js 1:17-61
 @ ./src/vendor.browser.ts
```
Maybe you can help me.

shlomiassaf:
Looking at the current state of the repo some changed were made the broke the github pages feature.

For example, the `metadata` property on the config object is now gone and it's part of the `HtmlWebpackPlugin`, we need it so we know the `baseUrl`.

The `output.publicPath` is also gone (webpack updates) which is also important for url purposes.

I will take a look and refactor it

shlomiassaf:
@katallaxie , it looks like [your commit](https://github.com/AngularClass/angular2-webpack-starter/commit/7bfba744ebfa77c5dfb30524e639e394a1d92c0e#diff-f737fedb5357468044db2dbe98eb0218L39) move the `metadata` property from the root of the config object into the  `HtmlWebpackPlugin` in `webpack.common.js` however other features relay on it for useful information (i.e: it's not only for the HtmlWebpackPlugin). Github pages is one but also [webpack.dev.js#23](https://github.com/AngularClass/angular2-webpack-starter/blob/master/config/webpack.dev.js#L23)  and [webpack.prod.js#27](https://github.com/AngularClass/angular2-webpack-starter/blob/master/config/webpack.prod.js#L27) assume it's there.

I will duplicate the `metadata` property in `webpack.common.js` to be both on root object and in `HtmlWebpackPlugin` for now to restore order.

However, some refactoring is needed since the logical process is wrong.
`HtmlWebpackPlugin` is set in `webpack.common.js` but a user might want to alter some values in dev/prod... 

It's not for this topic but I think common should have a pre process callback that can get the (almost) final config object and do some work such as adding plugins etc using updated/final objects.

shlomiassaf:
Here's a great example of why we need to pre process in common.

The GitHub pages feature requires some changes in the `HtmlWebpackPlugin.
Since GitHub pages sites have a URL path prefixed with the repo's name it must be set as the base url (e.g: angularclass.github.io/**angular2-webpack-starter**). Most (if not all) projects does not care about GitHub pages so the configuration is usually set so the app is in the domain's root (e.g: www.example.com/).

the GitHub pages feature must be transparent so it need to be able to set the configuration values of `HtmlWebpackPlugin` which is not possible (cleanly) at the moment.

shlomiassaf:
OK, issue fixed without changing `webpack.common.js`
However, relaying on HtmlWebpackPlugin API 

Also added a workaround for a bug in GH Pages after their move to Jekyll 3.3, this workaround must be reverted by people who use Jekyll in their GH pages.

shlomiassaf:
@katallaxie please review before I merge.

shlomiassaf:
reminder, can I have another eye looking at this before I merge?

@katallaxie @gdi2290 

Thanks!

PR #1091, opened 2016-10-07
Embraces angular2 styleguide

dherges:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Follow up on #1084 

* **What is the current behavior?** (You can also link to an open issue here)

Spec file names, e.g. ``x-large.spec.ts``

Component class names e.g.  ``App``

* **What is the new behavior (if this is a feature change)?**

Spec file names, e.g. ``x-large.directive.spec.ts``

Component class names, e.g.  ``AppComponent``

* **Other information**:

See
https://angular.io/styleguide#!#02-03
https://angular.io/styleguide#!#02-10
https://angular.io/styleguide#!#02-11

Styleguide also suggests to name e2e spec files with pattern: ``<feature>.e2e-spec.ts``, e.g. ``app.e2e-spec.ts``

PR #1084, opened 2016-10-05
Embraces ng2 styleguide *.component.{css|html|ts}

dherges:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Embraces angular2 styleguide for naming of component files, see: https://angular.io/styleguide#!#file-structure-conventions

* **What is the current behavior?** (You can also link to an open issue here)

``*.css``, ``*.html``

* **What is the new behavior (if this is a feature change)?**

``*.component.css``, ``*.component.html``

* **Other information**:

PR #1105, opened 2016-10-15
Feature/happy lint

michaeljota:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

tslint update
- **What is the current behavior?** (You can also link to an open issue here)
  - Tslint 3.9.x have a bug while dealing with wilcards
  - Codelyzer rules have not been used
  - Recommended rules hard coded.
- **What is the new behavior (if this is a feature change)?**
  - TSlint update to `@next`
  - Recommended Codelyzer rules being used in a separate file.
  - Extend recommended rules.
  - Update all code to follow new rules
- **Other information**:

This somehow is related to #1091 and #1084. But instead of **just follow the style guide** I updated the TSLint to linter the style guide with `codelyzer`.


michaeljota:
Ok. I did not see #185 before this. BUT, there is now an official [Angular 2 Style Guide](https://angular.io/docs/ts/latest/guide/style-guide.html) there is even a TSLint plugin to use them, that by the way this seed "use", named [codelyzer](https://github.com/mgechev/codelyzer), with the apropiated rules, everything should be fine. 

I just wanted to help.

katallaxie:
I did merged #185 and we have to adjust some things.

michaeljota:
I saw it. But as I say, this is not just "styling" the seed to Angular2 style guide, but instead, linting it, and fix the files to the linting, to Codelyzer, that happens to be the plugin to follow Angular2 style guide.

michaeljota:
Now that you merge #1091 I guess this have better chances, since now it only config the linter to use recommended rules, from TSLint and Codelyzer.

michaeljota:
I could do that. I guess that it's better if it's clear where the rules come from, if there are coming from external plugin.

katallaxie:
@michaeljota I just thought the same, yet coming from a different angle. I was think that in this starter, a single point for the rules is the way to go. So, I would just keeping everything in `tslint.json`.

michaeljota:
I see your point. I have a question then, should the recommended rules still explicitly be available or should they be extended?

d3viant0ne:
Other dependencies do not support tslint 4.x, resubmit a PR when both Codelyzer & tslint-loader all support the newer version of tslint. Merging something for 4.x / @next will fail peerDeps breaking nodejs 4.x builds.

Issue #576, opened 2016-05-09
Are there any plans to incorporate Angular Universal?


louisscruz:
- **I'm submitting a ...**
  [ ] bug report
  [ ] feature request
  [X] question about the decisions made in the repository

Are there any plans to incorporate Angular Universal? For me (and I assume for many others), this is a coveted feature. Perhaps there's a way to replace the http-server with [this](https://universal.angular.io/quickstart/) express server setup?

If there is no intent to include Universal, then maybe someone (more knowledgable than myself) could write a wiki on it? I'll be making an attempt in the next few weeks. If I happen to figure it out by then, I'll be happy to write one.

asgerjensen:
well, I got it working. 

Using the universal-starter as an offset, I created a seperate webpack config for the server build. Important notes:

1) Dont use chunks (vendor/polyfills). Build everything as one entry.
2) In tsconfig.json, remove the "noEmitHelpers" entry, or you will get errors like

__declare is undefined.

3) merge the dependencies from universal-starter (package.json and typings.json)
4) Add seperate scripts to the package.json (copy the dev or test targets).

To run, you'd then have two terminals open, one doing "watch" to rebuild, and one running nodemon to run/restart the server 


I dont have the expertise to make it one single webpack config.

Krisa:
Would be happy to see that also as I've not been very successful so far to integrate Universal with this starter. Especially, importing "angular2-universal/polyfills" introduces some unexplainable bugs in my project.

asgerjensen:
My webpack-universal.json, placed in the config directory.


```/**
 * @author: @AngularClass
 */

const webpack = require('webpack');
const helpers = require('./helpers');
var path = require('path');

/*
 * Webpack Plugins
 */
// problem with copy-webpack-plugin
var CopyWebpackPlugin = (CopyWebpackPlugin = require('copy-webpack-plugin'), CopyWebpackPlugin.default || CopyWebpackPlugin);
const HtmlWebpackPlugin = require('html-webpack-plugin');
const ForkCheckerPlugin = require('awesome-typescript-loader').ForkCheckerPlugin;


const ENV = process.env.NODE_ENV = process.env.ENV = 'development';

/*
 * Webpack Constants
 */
const METADATA = {
  ENV: ENV,
  title: 'Angular2 Webpack Starter by @gdi2290 from @AngularClass',
  baseUrl: '/'
};

/*
 * Webpack configuration
 *
[...]

PR #1165, opened 2016-11-10
feat: add support for Angular2 Universal

asgerjensen:
Feature - Adds angular2-universal as build-target.

Not using Universal
https://github.com/AngularClass/angular2-webpack-starter/issues/576

New build-targets (watch:universal, server:universal, build:universal).

* **Other information**:
This is probably best put in a separate branch, as I had to downgrade to angular 2.1.0 in order to get around https://github.com/angular/universal/issues/606

Also, I made no effort to minimize code duplication, as i wanted to keep the main-line code intact, and only demonstrate how to set up a universal build of the starter.


yuanfeiz:
Is this ready to be used for the latest version(2.4.1)?

d3viant0ne:
There is already a universal starter, adding it here would add unnecessary bloat

PR #1167, opened 2016-11-10
Close script tags in html-elements-plugin

sdemurjian:

Bug fix

Script tags are not properly closed when included in the html-elements-plugin

Script tags are now properly closed

* **Other information**:

My solution is a bit messy and I'm not sure whether or not the original behavior was intended. Anyways this was useful for me since I wanted to include a library from a CDN and have it in the same configuration file as the CSS and fonts loaded from CDNs.


sdemurjian:
let me know if this is good. converted to es6 template literals.

PR #1177, opened 2016-11-15
fix(angular2.2): wait for angular before disabling debug info

clementprevot:
Fix a bug that appears with Angular2.2 when using a production bundle.

Currently, when launching a prod bundle, the app doesn't work because of an `Cannot convert undefined or null to Object` that occurs in the `disableDebugInfo` function of Angular2.2.
This function tries to delete `window.ng.provider`, but at the moment of the call of the `disabledDebugInfo` function, the `ng` property has not yet been created in the `window` object because Angular is not yet initialized.

We now wait for Angular to be initialized before disabling debug info and enabling production mode.
This way, the `disableDebugInfo` call succeed and the app doesn't crash anymore.

* **Other information**:

### DIFF
--- a/src/app/environment.ts
+++ b/src/app/environment.ts
@@ -14,8 +14,12 @@ let _decorateModuleRef = function identity<T>(value: T): T { return value; };
 
 if ('production' === ENV) {
   // Production
-  disableDebugTools();
-  enableProdMode();
+  _decorateModuleRef = (modRef: any) => {
+    disableDebugTools();
+    enableProdMode();
+    
+    return modRef;
+  };
 
   PROVIDERS = [
     ...PROVIDERS,
--- a/src/app/environment.ts
+++ b/src/app/environment.ts
@@ -13,10 +13,11 @@ let PROVIDERS: any[] = [
 let _decorateModuleRef = function identity<T>(value: T): T { return value; };
 
 if ('production' === ENV) {
+  enableProdMode();
+  
   // Production
   _decorateModuleRef = (modRef: any) => {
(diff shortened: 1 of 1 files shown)


JihadMotii-REISys:
+1

This exactly what I was looking for 👍 .

Btw, I noticed another error. see attached
screen shot 2016-11-15 at 6 07 06 pm

clementprevot:
I didn't stumble upon the error you get. I'll dig a bit to see what's going on and let you know.
In the meantime, if you have any clue, let me know too.

clementprevot:
According to http://stackoverflow.com/questions/35721206/how-to-enable-production-mode-in-angular-2, `enableProdMode` should be called before bootstraping.

Let me fix my PR

JihadMotii-REISys:
+1 Thanks for the fix.

Now the app does get rendered but with another error message
 
screen shot 2016-11-16 at 11 42 12 am

clementprevot:
That's weird, I don't have these errors on my app...

src/app/environment.ts
```diff
@@ -13,17 +13,22 @@ let PROVIDERS: any[] = [
...
+  };
 
   PROVIDERS = [
     ...PROVIDERS,
     // custom providers in production
   ];
 
 } else {
-
+
```
d3viant0ne: @clementprevot - Please kill the whitespace with fire & rebase this with current master so I can get it landed.

clementprevot: Done

d3viant0ne:
Thanks @clementprevot

PR #1216, opened 2016-11-27
Update package.json

vadimzham:
Update

* **Other information**:
Updating package versions to be up to date


d3viant0ne:
@vadimzham - Thanks but we have greenkeeper.io for this. Also worth nothing more than a few of the packages you updated break functionality that will require additional work to upgrade.

PR #1226, opened 2016-11-30
chore(yarn): add yarn files to gitignore

d3viant0ne:
Ignores yarn output 

* **Other information**:

Per https://github.com/AngularClass/angular2-webpack-starter/commit/5db20a697bee643758ce597c8b9f7059f398e462#commitcomment-20016692

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,10 @@ Thumbs.db
 /bower_components
 npm-debug.log
 
+# Yarn #
+yarn.lock
+yarn-debug.log
+
 # Coverage #
 /coverage/
 


katallaxie:
See https://github.com/yarnpkg/yarn/issues/1583 that is a bad idea, when it comes to really use `yarn`.

d3viant0ne:
@katallaxie - The yarn.lock was removed by @gdi2290. The ignore for yarn is also at his request.

karolmie1:
@d3viant0ne Yeah It is really bad idea. yarn.lock is like one of the most important reasons to use yarn. 
I very often had this problem - I've downloaded old repo from github and I started to fire it up, but I couldn't manage to get it working, because author used older version of npm packages and the project won't compile anymore. (this also happend to me on material2 branch here if I remember correctly)

yarn.lock is very important and essential.

It may make pull requests and diffs ugly though.
Worry not, there is better solution than to remove it altogeter!

.gitattributes:
yarn.lock -diff

yarn.lock will be treated as binary from now on, won't be shown in difs and counted in lines changed if you care about such statistics.

katallaxie:
@d3viant0ne yep! you can remove it, but you should not `.gitignore` it, as people really use it in their projects.

karolmie1:
@katallaxie I will create a pull request with .gitattributes solution in a moment, you can check if you like it.

gdi2290:
yeah I didn't see that with yarn. let's just keep it out of the initial repo

irfannnn1:
,m

PR #1270, opened 2016-12-13
refactor sourceMapFilename to use file

gilhanan:
It's correcting to use [file] instead [name] in order to support work together with ExtractTextPlugin souremaps

Inspired by:
https://github.com/webpack/extract-text-webpack-plugin/issues/119#issuecomment-181584004


gdi2290:
SGTM

PR #1298, opened 2016-12-24
feature: added AOT compile support (#885)

### LINKED ISSUE #885: Ahead of time (AOT) compilation support
**Note: for support questions, please use one of these channels:** [Chat: AngularClass.slack](http://angularclass.com/member-join/) or [Twitter: @AngularClass](https://twitter.com/AngularClass)
- **I'm submitting a ...**
  [ ] bug report
  [x ] feature request
- **Do you want to request a _feature_ or report a _bug_?**
  Request a feature, AOT support in webpack
- **What is the current behavior?**
  JIT compile, performance is bad in prod.
- **What is the motivation / use case for changing the behavior?**
  Performance boost

colinskow:
This PR adds AoT support using the `ngc` compiler. `@ngtools/webpack` still has many flaws, but this setup is working perfectly. `npm install` and then `npm run ci` to fully test both AoT and JIT builds.

- upgraded Angular to ~2.4.1
- added AoT compile script
- fixed AoT incompatible code in app
- enabled Webpack 2 tree shaking by specifying ES2015 module output
- e2e testing script launches its own server
- ci script will test both AoT and JIT builds, makes fresh builds on each test
- specified `noEmit` typescript flag to prevent accidentally spamming compiled files into the project
- disabled caching on production server to be able to see changes
- updated bad links at top of README
- documented new features in README

Resolves #885


samvloeberghs:
wicked!

d3viant0ne:
@colinskow - I'll take a look at this in the morning.

//cc @gdi2290 @katallaxie should also take a look

d3viant0ne:
FYI - I restarted both the travis builds

.travis.yml
```diff
@@ -24,9 +24,4 @@ before_script:
   - sh -e /etc/init.d/xvfb start
   - sleep 3
 script:
-  - npm run lint
-  - npm test
-  - npm run build:prod
-  - npm run server:prod:ci &
-  - sleep 3
-  - npm run e2e
+  - npm run ci
```
d3viant0ne: I would also suggest running the two build / test steps individually. 

1.) Most of us run multiple test containers, I always run my jit / aot validations in parallel.
2.) If is step is going to fail, it's probably going to be AoT, splitting them just means less log output to go dig through to find the failure.

.gitignore
```diff
@@ -34,6 +34,7 @@ Thumbs.db
 /node_modules
 /bower_components
 npm-debug.log
+yarn.lock
```
d3viant0ne: The `yarn.lock` is set in .gitattributes as a binary file and shouldn't be ignored. See notes in #1226

colinskow: OK, I will remove this.

.travis.yml
colinskow: The complication in running the builds in parallel is that we need two different build directories and setup servers on two different ports to not cause conflicts. With this PR each step is an individual command and `npm run ci` simply composes them. I think parallel builds is a separate PR because it would take some work to setup.

Do you think the Travis builds would be more likely to pass the first time if I added a `sleep 3` between starting the server and starting the test?

colinskow: Also, after I got AoT working I hit a hitch where Angular was crashing on startup with the production build, even though the build was passing. That's why it is important to test both for this repo's CI. Individual developers can use `npm run ci:aot` for lint, karma, and test of AoT build.

colinskow:
@d3viant0ne `tsconfig.webpack.json` is correctly named. This is used by `awesome-typescript-loader` to run dev, production, and aot builds. It uses the es2015 module format in order to enable tree shaking.

Karma and other node libraries can't deal with es2015 modules and must use commonjs. That is why I forked the file. One for Webpack's internal use. The other for everything else.

.travis.yml
d3viant0ne: I personally use Circle but this should still be applicable ..

- I agree completely in regards to testing both builds, the point i was trying to make is to separate the two completely to allow the parallel execution. I have multiple containers available for any given build so my Jit & AoT validations run in parallel. Given it's a devops thing and most people aren't going to be running the vanilla Travis file for their projects, no need to have the build config hold up this PR.

- As far as the e2e tests go, I run a `sleep 5` after I spin up the server. I've had problems with 2-3 second sleeps when it's build prime time on Travis.

d3viant0ne:
re: `json config` misinterpreted the intent, makes sense.

.travis.yml
colinskow: @d3viant0ne are you talking about Docker containers? If you can get the builds to test in parallel and create a script command that would be great.

colinskow:
I've got another incoming commit on this PR to reduce bundle size and enable automatic tree shaking of the vendors bundle.

colinskow:
The AoT build was including lots of unnecessary files so I added a commit which automatically extracts imports from `node_modules` into the vendors module and tree shakes them. This cuts the output bundle size in less than half. Previously there was no tree shaking of vendor imports.

I also added the Webpack Bundle Analyzer plugin and made it accessible by setting the command line flag `--env.analyze`. This allows for easy inspection of output bundles to ensure they are fully optimized. For example: `npm run build:aot:prod -- --env.analyze`.

tsawan:
eagerly waiting for this PR to be merged.

Phmager:
Awesome work, you should also add a script for: npm run start:aot. 
And how can one import bootstrap 4 with that changed logic for referencing vendors?

colinskow:
@katallaxie not all Angular apps are compatible with AoT so I wouldn't want to break people's builds by setting it as default. The idea is to add features without making breaking changes.

As mentioned above @ngtools/webpack is still unstable and I wasn't able to get it to work with this repo. But when they work out the bugs it is the best way to go with Webpack.

The BundleAnalyzerPlugin adds 5.5mb to dependencies that are already 320mb+. It is very useful for ensuring that the builds contain what we expect them to contain and is optional to use via a command line flag.

colinskow:
@Phmager AoT is more for production builds. Since the NGC compiler is running separately from Webpack it may not play well with the dev server.

The new vendor referencing logic is that anything from `node_modules` that you import into your project will be automatically pulled into the vendor bundle. You don't have to do anything. For Bootstrap I would import at the top of your main app module file.

However its bad practice to have large CSS files inlined in the Javascript code, so my next step is a PR to setup a `styles` directory and everything in there will be compiled to a separate .css file.

d3viant0ne:
@katallaxie - AOT should not be the default for `prod` for exactly what @colinskow stated, it breaks things. After a developer has his application running stable on AoT, he or she can default it to AoT in their project.

Given the commitment required to create a large application published AoT, this is a feature that should be off by default & opt in or it's going to break every production build this starter backs when they pick up the changes.

d3viant0ne:
I do on the other hand agree about @katallaxie's comment in regards to the BundleAnalyzerPlugin, It's trivial to add and would suggest this be transfered to Wiki instructions & not added to the seed directly ( it's big ).

We get basic bundle analysis via Webpack RC, if people want all the bells & whistles they can add `BundleAnalyzerPlugin`

katallaxie:
@colinskow I get your point. Though, I believe it should be considered best practice to have AoT and deviate from that. So, if you have an application, that for what ever reasons, doesn't work with AoT then you can disable it. But I would see it the other way around. As regarding to `@ngtools/webpack`, I had doubts some months ago, that`s why it wasn't in the `aot` branch. Though, I see it as the tool of choice in the long run. You have less boilerplate code, better integration with what Webpack does and better code quality.

katallaxie:
@d3viant0ne Because we had some deps fixed, we should take care of them now. We should also update `webpack` to `2.2.0-rc.2` and `webpack-dev-server` to `2.2.0-rc.0`.

maxisam:
do we wanna update readme because the vendor file is gone ?

config/webpack.common.js
```diff
@@ -178,6 +179,17 @@ module.exports = function (options) {
        * See: https://github.com/webpack/docs/wiki/optimization#multi-page-app
        */
       new CommonsChunkPlugin({
```
albv: I don't get what this instance of CommonsChunkPlugin is supposed to do.. As far as I can tell, with this settings it won't do anything at all. @colinskow could you elaborate please?

abbilovos:
- [ ] ![@Uploading 154770969...] @

PR #1299, opened 2016-12-24
(fix) enable sourcemaps in production build

colinskow:
Bug fix: sourcemaps are no longer enabled by default in the UglifyJS Webpack plugin.

No source maps are generated in a production build.

Source maps are correctly generated.

* **Other information**:

### DIFF
--- a/config/webpack.prod.js
+++ b/config/webpack.prod.js
@@ -171,6 +171,7 @@ module.exports = function (env) {
           join_vars: true,
           negate_iife: false // we need this for lazy v8
         },
+        sourceMap: true
       }),
 
       /**


d3viant0ne:
@colinskow - See #1241

This is not something that should be on by default but opt in should a developer choose or need to push production sourcemaps.

d3viant0ne:
As an alternative, An off by default configurable flag would be fine if the idea is to not have to deal with overwriting the config setting. i.e `--sourcemaps` on the build call like env.

argasek:
Just to help other developers who pull their hair out trying to get the source maps on production build: this pull request is not sufficient anymore, as recent updates to OptimizeJsPlugin, which is included by angular-starter broke SourceMaps generation: https://github.com/vigneshshanmugam/optimize-js-plugin/issues/7

colinskow:
Developers that need sourcemaps in a production build should probably disable the optimize-js plugin then.

SherlockStd:
Thanks !

PR #1300, opened 2016-12-25
fix: Rename export for detail module

stevenjob:
Just a rename

### DIFF
--- a/src/app/+detail/index.ts
+++ b/src/app/+detail/index.ts
@@ -22,6 +22,6 @@ export const routes = [
     RouterModule.forChild(routes),
   ]
 })
-export class AboutModule {
+export class DetailModule {
   static routes = routes;
 }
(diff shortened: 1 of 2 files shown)


colinskow:
Also default needs to be removed because it is incompatible with the AoT compiler.
https://github.com/AngularClass/angular2-webpack-starter/pull/1298/files#diff-15e96909d26305ab86756a9d68100c27L25

I agree that `DetailModule` is a better name. If the maintainers agree I can add a commit to my PR. The link to the module in `app.routes.ts` also needs to be updated.

d3viant0ne:
@stevenjob / @colinskow - I'll merge yours as it is, then Steven can rebase & rename the component as discussed above.

katallaxie:
👍

stevenjob:
Done 👍

PR #1318, opened 2016-12-31
.spec examples for unit testing

katallaxie:

Changes the unit test implementation as to be more compliant with the Angular 2 Docs. Needs further discussion.

Not thus compliant unit testing.

Compliant unit testing.

* **Other information**:


katallaxie:
We should work on that, as seen in the one example.

PR #1341, opened 2017-01-06
(enhancement) use DLLs to speed up dev builds

colinskow:
This PR introduces DLLs to dramatically speed up development builds. This theoretically should have zero side effects and not break any existing builds. (Although we should test thoroughly in the wild before merging.) The `check-dlls` script automatically detects when DLLs are missing or out of date (when the installed packages they reference change versions) and rebuilds them only when needed.

These enhancements are completely transparent to the developer. The user doesn't have to do anything to get performance benefits, but can edit `config/dll.config.js` to add additional packages to the DLL build. Anything that is missing will end up in common chunks and not cause any breakage (but won't be as fast as it could be).

In addition I:
- Added the DLL dev build to the `npm run ci` test so we don't break it by accident in the future
- Cleaned up the NPM scripts and removed some redundant ones that were getting called twice during builds
- Updated the README with DLL info and instructions for external stylesheets

So far I've tested this PR on MacOS Sierra and Windows 10.


abhishekisnot:
Not sure how much time we would save on polyfills DLL alone. Is it worth adding polyfills DLL to prod build? I suppose AOT does not look at polyfills.

d3viant0ne:
No, it doesn't. You are never going to see a real time savings in the starter or most of the applications built on top of it less the very largest applications. But hey, the masses just had to have the next hot buzz word, regardless of their understanding of it's benefits and @colinskow was kind enough to provide.

colinskow:
On my MacBook Pro I'm getting 11,003ms dev build time for master, and 6345ms for DLL branch dev build. That's almost 5 seconds faster or a 40% gain. This isn't much time on one build, but when a developer is repeatedly testing code changes in watch mode it adds up quick.

I don't think it will provide any benefits to production builds though.

colinskow:
@shlomiassaf's implementation is better so I'm closing this ;-)

PR #1343, opened 2017-01-06
Update tsconfig.webpack.json

maxisam:
According to https://www.typescriptlang.org/docs/handbook/module-resolution.html#base-url
It should be for non-relative modules

bug fix ? It probably doesn't really cost anything since we set  "moduleResolution" as "node"

* **Other information**:

### DIFF
--- a/tsconfig.webpack.json
+++ b/tsconfig.webpack.json
@@ -10,7 +10,7 @@
     "noEmit": true,
     "noEmitHelpers": true,
     "strictNullChecks": false,
-    "baseUrl": "./src",
+    "baseUrl": "./node_modules",
     "paths": {
     },
     "lib": [
--- a/tsconfig.webpack.json
+++ b/tsconfig.webpack.json
@@ -10,9 +10,6 @@
     "noEmit": true,
     "noEmitHelpers": true,
     "strictNullChecks": false,
-    "baseUrl": "./node_modules",
-    "paths": {
-    },
     "lib": [
       "es2015",
       "dom"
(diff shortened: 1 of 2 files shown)


d3viant0ne:
As I understand it and i'm going off docs I read quite some time ago, we only need to set `baseUrl` if we are defining `paths` which iirc is just set to an empty object.

//cc @colinskow @gdi2290 <--- Need someone more knowledgable with the inner workings of TypeScript to answer this one :)

maxisam:
Yes, I would say either remove it or set it to ./node_module, set it to ./src is actually misleading.

colinskow:
When I was working on AoT I didn't know if `"baseUrl": "./src"` served a purpose so I just left it. I do believe that `node_modules` IS by default a base URL for TypeScript. So if the CI tests pass we could remove it from all the `tsconfig` files.

d3viant0ne:
@maxisam - Given we are not actually using `paths` anywhere, would you please update this to pull both `paths` & `baseUrl` from all the typescript configuration files?

maxisam:
ok it is done.

tsconfig.webpack.json
```diff
@@ -33,10 +30,12 @@
     "useWebpackText": true
   },
   "angularCompilerOptions": {
-   "genDir": "./compiled",
-   "skipMetadataEmit": true
+    "genDir": "./compiled",
```
colinskow: Indentation mistake

colinskow:
There's a tiny indentation mistake introduced, which I pointed out. Otherwise looks good to me.

tsconfig.webpack.json
maxisam: I am confused. I thought indentation should be like 2 spaces ?

colinskow: OK, now I see. It was wrong and you corrected it. Never mind!

Issue #1345, opened 2017-01-06
Feature - Autoprefixer Support


abhishekisnot:
* **I'm submitting a ...**
[ ] bug report
[x] feature request
[ ] question about the decisions made in the repository

* **Do you want to request a *feature* or report a *bug*?**
*feature*

* **What is the current behavior?**
Both global and component stylesheets do not have vendor prefixes to CSS

* **What is the expected behavior?**
Both global and component stylesheets should have vendor prefixes to CSS

* **What is the motivation / use case for changing the behavior?**
It can often become a bit tedious to include all the vendor prefixes for the CSS you use. Use an auto-prefixer to automatically add the prefixes you need to include.

* **Other information**
[autoprefixer](https://github.com/postcss/autoprefixer) is a PostCSS plugin to parse CSS and add vendor prefixes to CSS rules using values from [Can I Use](http://caniuse.com/). It is [recommended](https://developers.google.com/web/tools/setup/setup-buildtools#dont_trip_up_with_vendor_prefixes) by Google and used in Twitter and Taobao.

abhishekisnot:
Refer #1345 for discussion on this.

PR #1347, opened 2017-01-06
Feature #1345 - add autoprefixer support using postcss

abhishekisnot:
*feature*

Feature #1345 

AutoPrefixer will add vendor prefixes after this PR is merged.

* **Other information**:
> The `*.scss` loader was missing from `webpack.test.js` which was causing failures if one had scss file in a component.
> It's probably better to add [null-loader](https://github.com/webpack/null-loader) for `*.css` and `*.scss` in test configuration. Maybe we can create a separate PR for that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ Thumbs.db
 /bower_components/
 npm-debug.log
 /npm-debug.log.*
+yarn.lock
 
 # Coverage #
 /coverage/
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -4,6 +4,7 @@
 
 const webpack = require('webpack');
 const helpers = require('./helpers');
+const autoprefixer = require('autoprefixer');
 
 /*
  * Webpack Plugins
@@ -134,7 +135,7 @@ module.exports = function (options) {
          */
         {
           test: /\.css$/,
-          use: ['to-string-loader', 'css-loader'],
+          use: ['to-string-loader', 'css-loader', 'postcss-loader'],
           exclude: [helpers.root('src', 'styles')]
         },
 
@@ -145,7 +146,7 @@ module.exports = function (options) {
          */
         {
           test: /\.scss$/,
-          use: ['to-string-loader', 'css-loader', 'sass-loader'],
+          use: ['to-string-loader', 'css-loader', 'postcss-loader', 'sass-loader'],
           exclude: [helpers.root('src', 'styles')]
         },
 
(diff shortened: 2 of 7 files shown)


colinskow:
I think the question to ask is how many developers need this. Most modern browsers support just about all CSS tags without prefix now. This is not needed for Safari, Chrome, Firefox and IE10+.

This is going to slow the build slightly and add size to compiled CSS. The purpose of this repo is to provide a starting point for majority use-cases. This is an easy modification to make for those who have a specific need for it.

d3viant0ne:
I am pretty sure the stance on autoprefixer in the past was if people want to use it, they can add it but @gdi2290 didn't want it in the starter as it adds more complexity for limited usefulness.

Personally I agree with @colinskow in regards to this being a starting point and thus we really only want to include things that are of benefit to everyone using the starter to build upon.

shlomiassaf:
We also need to keep in mind that every new feature adds regression debt.

IMO, the purpose of this starter is to provide the best build process, that does the essentials, in zero time.
It also needs to include some demo's for main features like lazy loading, injection etc...

Now, I can live with SCSS since it is considered essential but any post css out there is not.

I think the rule of thumb should be, will it work without it?
Not having a SCSS loader means SCSS will not work, so include it.
Not having an auto prefixer, still everything works.. dont include it.

Next thing will be adding translation support just cause it's cool :)

Again, IMO, This is the same approach we need to take with DLLs, its not essential nor expected by developers, it's still an "early" concept and make entry level users run away...
Yep, it's cool to have I agree, but this is a highly used repository so people should understand what's happening.

colinskow:
@shlomiassaf just to be clear the case for the DLLs pull request is that it cuts dev build time in half, which leads to dramatic productivity gains -- especially when repeatedly testing small changes in code or debugging. It's a money maker for the companies who will use this starter.

The DLLs are setup so they are totally transparent to the user, who gets automatic immediate productivity gains as soon as we test and merge the PR. It's not going to scare new people away since they don't need to get into technical details (which is just one plugin added to the dev build chain). It also shouldn't break any existing builds.

katallaxie:
@shlomiassaf I could live with Autoprefixer + PostCSS, rather then SCSS 🤕 . There are many projects that use different CSS preprocessors. However, there are litte differences at to what is need in the post process, which is compression, optimization, and QA.

d3viant0ne:
We are going to stay the course here. If a developer wants / needs PostCSS / Autoprefixer they can add it to their project. 

Closes ##1345

Issue #1364, opened 2017-01-11
watchOptions - webpack have to be changed.


czegarram:
Lines:

`      watchOptions: {
        aggregateTimeout: 300,
        poll: 1000
      }`

have to be replaced by:

`      watchOptions: {
        aggregateTimeout: 300,
        poll: false
      }`

In file: (line 220)

[Webpack Dev](https://github.com/AngularClass/angular2-webpack-starter/blob/master/config/webpack.dev.js)

Because is generating a cpu problem in development environments that have Windows as Operating System.  

My specs:

- Windows 7
- Intel core i7 6600U 2.80
- 8GB Ram.

Cpu has 95% of use when i run 'npm start'. But, when I change that configuration option, it seems to start to work correctly.

colinskow:
This is a good idea. `poll: false` won't work inside VirtualBox and a few other situations, but we can leave a commented note for those folk. This will improve performance for everybody else. In addition we could probably further improve watch performance with `ignored: /node_modules/`.

https://webpack.js.org/configuration/watch/

d3viant0ne:
@colinskow May as well roll this into the webpack config rework and make it a configurable option for the user.

I'll rework them later this evening

colinskow:
@d3viant0ne definitely add `ignored: /node_modules/` since that will definitely improve performance on lower end systems for everyone.

colinskow:
I discovered an issue with `awesome-typescript-loader` that is related to this and contributing to poor performance. It is also preventing me from being able to separately watch and reload the Main and Renderer processes in my Electron Branch.

https://github.com/s-panferov/awesome-typescript-loader/issues/351

PR #1378, opened 2017-01-16
Enhancement: TS compiles only dependencies instead of entire project

### LINKED ISSUE #1366: Should compiled/ directory be excluded by at-loader
* **I'm submitting a ...**
[ x] bug report
[] feature request

* **Do you want to request a *feature* or report a *bug*?**
bug  - but maybe a feature :-)

Currently if you have run an aot build and have a compiled/ folder, and then run the dev server, then at-loader runs across the compiled directory also, which increases compile time by a couple of seconds, but also shows errors if constructor injections in src folder change. 

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** 

With a fresh copy of repo run the start:hmr and check at-loader completion times, then run build:aot, and then start:hmr again - at-loader times possible increased by several seconds.

If you delete compiled folder or change excludes in tsconfig.webpack.json to the following then at-loader completion times go down again.

  "exclude": [
    "node_modules",
    "dist",
    "compiled",
    "src/**/*.spec.ts",
    "src/**/*.e2e.ts"
  ]

The above change does not appear to affect the AoT build process at all.
Maybe cleaning the compiled folder at the start of the dev server task is a better option?

* **What is the expected behavior?**

Existence of the compiled folder shouldn't affect dev build at all?

* **What is the motivation / use case for changing the behavior?**

The extra at-loader completion time is not such a big issue, but getting at-loader errors about code in the compiled folder when running dev task is annoying.

* **Please tell us about your environment:**
- Angular version: 2.4.1
Windows

colinskow:
This PR improves the efficiency of the TypeScript build and resolves several issues that result from TypeScript compiling unnecessary files.

The current behavior is that on every build TypeScript compiles every single `.ts` file in the project tree. This is inefficient and can lead to issues when different files in the project require different configurations or use conflicting Typings.

The solution is to tell TypeScript specifically which entry files to compile, and it will automatically compile all their needed dependencies too. With this configuration TypeScript avoids compiling unnecessary files. `awesome-typescript-loader` doesn't seem to support passing in a configuration object, so this PR introduces a Webpack plugin which writes a temporary `tsconfig.json` file containing the correct entry files.

The only drawback is that lazy loaded modules need to be declared in `src/app/lazy-loaded.ts` or they will not get compiled. Maybe @shlomiassaf can write a script that automatically detects and adds them.

- Resolves #1366 
- Improves #1364 

Overall this PR makes builds faster, avoids some Typings conflict issues, and makes watch mode more efficient by watching only necessary files. It is also needed for my Electron branch to separately watch the main and renderer builds.


colinskow:
Looks like Awesome TypeScript Loader is going to make this much easier to implement, so might as well wait until this issue is resolved before merging.
https://github.com/s-panferov/awesome-typescript-loader/issues/364

shlomiassaf:
@colinskow This introduces a lot of pitfalls.

The main one is the need to specify all of your lazy modules in a special files.
It makes lazy loading difficult more than it already is... lazy loading should be part of the route definition flow without extra requirements... 
This alone is a blocker for me...

Also, the final tsconfig is not known to the user, very blurry...

I think it will create more confusion

colinskow:
@shlomiassaf 

There are two ways TypeScript can work with lazy loaded modules:
1) Compile every single file in the project
2) Compile entry files, plus declared lazy loaded modules

If you want to avoid compiling everything, then there is no way around declaring lazy loaded modules. Remember when we used to have to declare vendor libraries?

The way around this would be to write a script that parses the dependency tree and detects lazy loaded modules on its own. But that's a lot of work for very little productivity gain in my eyes.

shlomiassaf:
@colinskow I'm not sure you are right, if this is the case lazy routes would never have worked and this is exactly why `ng-router-loader` exists.

Although `ng-router-loader` kicks in after `awesome-typescript-loader` finished TS to JS conversion it still adds a reference to a file (which is TS).
Webpack will load this file and send it through TS loader chain.

So anyway, the files goes through TS compilation hence TS type checking.

If `awesome-typescript-loader` will typecheck everything in the path of the entryPoint we're good.

shlomiassaf:
For example, `ng-router-loader` will convert this:
```
 { path: 'detail', loadChildren: './+detail#DetailModule'},
```

into this:
```ts
function() { return import('/forks/angular2-webpack-starter/src/app/+detail/index')  .then( function(module) { return module['DetailModule']; } ); }
```

Now this is pure JS code at this point but webpack will parse it to see if there are any `require` statements or `import()` in this case.

Since there are it will load them, since this points to a TS file it will go through the TS chain... all good.

gdi2290:
@shlomiassaf @colinskow any updates on this PR? I really want to merge it in

colinskow:
@shlomiassaf the lazy loaded files are compiled because TypeScript's default behavior is to compile EVERY SINGLE FILE in the project tree. If we tell ATL to only compile entry files and their dependencies, it will skip the lazy loaded files unless they are declared explicitly.

This PR adds a bit of complexity to the build process, but also makes it go much quicker if there are a lot of unused TS files in the tree. (For example e2e and unit tests.)

It is absolutely necessary in my [Electron fork](https://github.com/colinskow/angular-electron-dream-starter) since I have different builds for the main process, renderer process, Karma tests and Spectron tests, and the types conflict.

In this project it would probably give a couple second boost to each build, but nothing huge. The biggest benefit is you don't have to worry about type conflicts.

Two options:
1) Declare lazy loaded modules in a separate file (existing behavior)
2) Write a loader which resolves lazy loaded modules prior to ATL running

Also still waiting on https://github.com/s-panferov/awesome-typescript-loader/issues/364 which will remove the need to write temporary .tsconfig files.

PR #1382, opened 2017-01-16
update ngc-webpack to latest

shlomiassaf:
This upgrades to ngc-webpack 1.1.2

In 1.1.0 depends on `@angular/tsc-wrapper` and `@angular/compiler-cli`

When `@angular/compiler-cli` bumps `@angular/tsc-wrapper` then **ngc-webpack** needs to do so as well... which is not ideal.

`@angular/compiler-cli` actually is `@angular/tsc-wrapper` since it re-exports all of the tsc-wrapper api... 

1.1.2 uses `@angular/compiler-cli` only.

I had a break when trying to compile a complex setup, took me 1 hour to find the culprit, `@angular/compiler-cli` was using `@angular/tsc-wrapper` 0.5.1 while **ngc-webpack** was using 0.5.0 :/

**Note that for some reason the tests fail on NODE 5, I don't know why yet... so this version is 6+**
The error:
```
  AssertionError: expected [Error: undefined, resolving symbol OpaqueToken in /home/travis/build/shlomiassaf/ngc-webpack/node_modules/@angular/core/src/di/opaque_token.d.ts] to be undefined
```

The build using 2 steps (ngc-w -> webpack) from the cli works!
Just from the plugin it doesn't very strange...

> I'm not sure it's not working on 5, my demo app for integration test might be the issue or it just doesn't work...

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "karma-sourcemap-loader": "^0.3.7",
     "karma-webpack": "2.0.1",
     "ng-router-loader": "^1.0.2",
-    "ngc-webpack": "1.1.0",
+    "ngc-webpack": "~1.1.2",
     "node-sass": "^4.2.0",
     "npm-run-all": "^4.0.0",
     "parse5": "^3.0.1",


shlomiassaf:
OK, also fails here on NODE 5.

```
Angular compilation done, starting webpack bundling.
Error: undefined, resolving symbol OpaqueToken in /home/travis/build/AngularClass/angular2-webpack-starter/node_modules/@angular/core/src/di/opaque_token.d.ts
```

Any clues on why?

d3viant0ne:
Given that webpack doesn't support Node 5, does it really matter?

shlomiassaf:
IMO, no, we should support LTS ...

But you know how it goes... :)

colinskow:
I think we need to add a deprecation notice for Node 5 support to the top of the docs right away and ask people to upgrade to the latest version of Node 6 LTS as soon as possible. Otherwise we can't advance our dependencies anymore.

shlomiassaf:
@gdi2290 @d3viant0ne @colinskow 

We need to decide on this one, if we won't upgrade we will soon start to see issues flying around....

colinskow:
Let's ditch Node 5... It was never LTS and was a short-lived version. Our dependencies don't support it anymore, so why should we?

1) Update the docs stating Node 6 LTS is the minimum
2) Remove Node 4 altogether from Travis, and replace Node 5 as the allowed failure

If @d3viant0ne agrees let's pull the trigger.

gdi2290:
@shlomiassaf you can require the version @angular/compiler-cli/package.json and require the correct file for each version to support any version

shlomiassaf:
@gdi2290 I don't follow...

We need to decide on this, `ngc-webpack` 1.2 has some bug fixes and ability to pass emitted files (assets) back to webpack (`file-loader`)... so we need to upgrade.

colinskow:
Thank you very much @shlomiassaf I updated to 2.0.0 in #1691

PR #1387, opened 2017-01-17
feat: use import() construct via ng-router-loader@2.0.0

shlomiassaf:
Support the [Dynamic import proposal](https://github.com/tc39/proposal-dynamic-import) [TC39 Stage 3] as a loader (codegen).
The import() construct is supported in webpack from 2.1.0-beta28

> "async-system" is deprecated in webpack 2 and removed in webpack 3.

Moving to 2.0.0 requires a subtle change in the TS loader order.

**Dev's that just upgrade the packages will have some trouble locating this change in the configs so we might expect some noise**

Reason for change:
Typescript transform `import(...)` syntax into something else.
To use `ng-router-loader` with `async-import` code generator the `ng-router-loader` must
 run **AFTER** the TS compiler (e.g: awesome-typescript-loader), that is in a lower index in the loaders array. If it runs before, TS will see the `import(...` expression and will convert it into invalid code.

This change also requires the code generators to emit ES5 code, all codegen's now emit ES5 code.

SEE: Microsoft/TypeScript#12364

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -93,17 +93,29 @@ module.exports = function (options) {
       rules: [
 
         /*
-         * Typescript loader support for .ts and Angular 2 async routes via .async.ts
-         * Replace templateUrl and stylesUrl with require()
+         * Typescript loader support for .ts
+         *
+         * Component Template/Style integration using `angular2-template-loader`
+         * Angular 2 lazy loading (async routes) via `ng-router-loader`
+         *
+         * `ng-router-loader` expects vanilla JavaScript code, not TypeScript code. This is why the
+         * order of the loader matter.
          *
          * See: https://github.com/s-panferov/awesome-typescript-loader
          * See: https://github.com/TheLarkInn/angular2-template-loader
+         * See: https://github.com/shlomiassaf/ng-router-loader
          */
         {
           test: /\.ts$/,
           use: [
-            '@angularclass/hmr-loader?pretty=' + !isProd + '&prod=' + isProd,
             {
+              loader: '@angularclass/hmr-loader',
+              options: {
+                pretty: !isProd,
+                prod: isProd
+              }
+            },
+            { // MAKE SURE TO CHAIN VANILLA JS CODE, I.E. TS COMPILATION OUTPUT.
               loader: 'ng-router-loader',
               options: {
                 loader: 'async-import',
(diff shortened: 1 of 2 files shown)


shlomiassaf:
@colinskow that's actually not related.

The modules are loaded at startup due to the route strategy, this is configurable.

They are still asynchronous, they just load automatically at startup but a separate bundle is there for each module.

This update doesn't add new functionality, it's more syntax related.

End result for all async code styles (require.ensure, System.import and import()) is logically the same...

colinskow:
How would you configure it to load the chunks on route change?

shlomiassaf:
https://github.com/AngularClass/angular2-webpack-starter/blob/master/src/app/app.module.ts#L63

Just remove the strategy the default should be nod preloading I think.

Anyway you can change it to what's available (see the docs) and you can create custom strategies

d3viant0ne:
Outside of Patrick's approval, we need to get decent documentation around this with all the other routing docs. Most of the users of this starter don't follow TypeScript issues / TC-39 like we do.

shlomiassaf:
Worth noting that in 2.0.0 you had to move the loader to the top only if you wanted to use `asyn-import`, that is, use the new  `import()` dynamic module loading construct.

In 2.1.0 it is required for **all** use cases since I switched to JS AST, if you sent TS to the loader it will fail... 

It doesn't make sense to use the TS compiler here since angular can run on raw JS code

shlomiassaf:
@d3viant0ne, last commit reasoning

The definition of the `*.ts` loader is getting more complex.
We need AOT compliance so we use `angular2-template-loader`
We need lazy loading so we use `ng-router-loader`
And of course, we need to compile TS to JS, `awesome-typescript-loader`

We can't explain all that in the configuration file but we can use Webpack 2 new loader syntax and use objects instead of literal string to configure the loaders.

shlomiassaf:
@d3viant0ne Please review https://github.com/AngularClass/angular2-webpack-starter/wiki/Routing-and-Lazy-loading-@NgModules

Comments are welcome

shlomiassaf:
@gdi2290 make a quick review on this one, it's a block for another PR and the current loader's version (2.1.0) has bug fixes and improvements that we need to push into the starter.

colinskow:
@shlomiassaf I made some documentation on lazy loading here:
https://github.com/AngularClass/angular2-webpack-starter/pull/1378/files#diff-04c6e90faac2675aa89e2176d2eec7d8R251

You could probably take the first part of that.

PR #1391, opened 2017-01-18
 fix(deps): allow to make a build without dev dependencies

DanielKucal:
Hello all, I use and really appreciate this seed, but there is one thing that keeps me awake at night... 

In this PR I've moved development dependencies that are mandatory to make a build (and only them) into normal dependencies.

For now, when someone tries to make a build in a production environment (`NPM_CONFIG_PRODUCTION`=`true`, `NODE_ENV`=`'production'`) dev dependencies are not downloaded while some of them are required to make a build. Developer will be getting many different errors, like:
>     Error: Cannot find module 'awesome-typescript-loader'
>     Cannot find type definition file for 'chai'.
>     sh: 1: webdriver-manager: not found
until he moves all needed deps or perchance installs them manually. It's time-wasting and the solutions are not always obvious (e.g. `webdriver-manager` command comes from `protractor` repository). 
With current behavior it is impossible to deploy application i.e. on Heroku.

* **What is the new behavior?**
After this change production (and development) build runs smothly.

I will be happy to have it merged and respected in the future or to see better solution. Thanks in advance!


colinskow:
My opinion is that dependencies are what is actually include in or used by the built app and devDependencies are all the tools for the build process. Generally you shouldn't have `NODE_ENV="production"` set when you `npm install` if you intend to make a build.

d3viant0ne:
This is a CI / CD issue that needs to be addressed in your deliver workflow.

`devDependencies` exist as to not bloat the size of the production application which at the end of the day is payload a user has to download. This is not something we would be interested in changing here and I can't stress to you how important it is you don't do this in your application.

DanielKucal:
So what solution would you recommend to deploy `angular2-webpack-starter`-based application i.e. on Heroku?

colinskow:
@DanielKucal 

Option 1) Pre-build and push the final product to Heroku
Option 2) `heroku config:set NPM_CONFIG_PRODUCTION=false`

https://devcenter.heroku.com/articles/nodejs-support#devdependencies

PR #1418, opened 2017-01-25
Use of v8-lazy-parse-webpack-plugin

killzoner:

Clean prod config / package.json

Currently the **v8-lazy-parse-webpack-plugin** is only referenced in this [file ](https://github.com/AngularClass/angular2-webpack-starter/blob/fa823c2ac46b2d87cf591937c8799b193bd5f216/config/webpack.prod.js#L20) but never used.

Should we either use it or remove dependency (PR subject)?

Remove unused plugin


katallaxie:
See #1419

katallaxie:
Closed #1419

PR #1432, opened 2017-01-27
use package.json:applicationName to control app title instead of config/webpack

seiyria:
Move application config from a random area in the config folder to the package.json file.

Title is hardcoded in the config folder

Title configuration has been moved to package.json

* **Other information**:
I found this particularly useful because I also used a bootstrap header bar with the title, and having it in two places was particularly annoying, so this was a good maintainability change for me.

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -4,6 +4,7 @@
 
 const webpack = require('webpack');
 const helpers = require('./helpers');
+const package = require('../package.json');
 
 /*
  * Webpack Plugins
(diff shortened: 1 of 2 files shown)


bmayen:
I would prefer to have project-specific config pulled out and merged in via separate configs. App title is just one example, but there are other things like the DLL config which don't make sense to add to package.json. Separate project config would cover all cases. 

I know this adds to complexity a bit, but it makes it easier to pull updates from this starter into existing projects.

seiyria:
Maybe, but I generally prefer to not have too many config files. A separate application block in package.json generally suffices for me. That said, this is pretty subjective. If it's not generally applicable enough it could be closed, but I still think any config-related data (not to be confused with the bootstrapping stuff in `config/`) should be pulled out.

bmayen:
Problem with an application block in package.json is that it complicates things when dealing with something other than strings. For instance, configuring loaders or plugins. If we did separate project configs, they could be very simple by default so as to not unnecessarily complicate/confuse things. Optional config if you need it, but otherwise you could use the starter as-is.

gdi2290:
we probably shouldn't include more metadata in package.json unless it's something a lot of developers do

PR #1438, opened 2017-01-29
refactor to place dev, prod, and aot builds under single 'build' folder

bmayen:
Feature/refactor

Both dev and prod builds build to a single 'dist' folder so it is not easy to build both and compare side by side. Additionally, when cleaning the project, this folder is removed which can be a problem in IDEs when trying to permanently exclude the folder from the project.

Builds are placed under a single "build" folder. This allows dev, prod, and aot builds to coexist. It also allows more targeted cleaning.

* **Other information**:


bmayen:
Not sure how to resolve this last conflict. Locally I'm able to merge without conflict now. And not keeping my changes would cause the feature in this PR to break.

bmayen:
I resolved the conflicts in this branch, but it did not trigger a new CI build. Could somebody with access please trigger one?

gdi2290:
this is a great idea. sorry I didn't get it to. I'll try to add it myself during the refactor of the repo

PR #1447, opened 2017-01-31
Use tslib instead of ts-helpers

JoniJnm:
Typescript 2.1 has a new option called `importHelpers` that allows use `noEmitHelpers` without `ts-helpers`.

In [ts-helpers wiki](https://github.com/ngParty/ts-helpers):

> NOTE: Starting Typescript 2.1 this package won't be needed anymore

More info:

* https://github.com/Microsoft/TypeScript/pull/9097
* https://github.com/Microsoft/tslib

With this change we can use `extends` in protractor ts files for e2e tests.

If you don't want do this change, you should add `require('ts-helpers');` at the beginning of the file `protractor.conf.js`.


katallaxie:
👍

PR #1491, opened 2017-02-07
Setup node mem directly in `package.json` script

elhigu:
Compiling AoT build uses huge amount of memory in any bigger project. Commit adds an example how to setup memory limits in `package.json`.

Feature for setting more memory available for webpack, since AoT compilation uses easily > 2-3GB

Node crashes to allocation errors when compiling bigger project and user has to start googling, why this might happen (and there wasn't good examples how to set value to scripts ran through npm).

* **Other information**:


pilgylis:
Brilliant!!

Finally got the AOT running!!

However I have to make the following changes to run on my machine (Windows 10):
"webpack-dev-server": "node --max_old_space_size=4096 node_modules/webpack-dev-server/bin/webpack-dev-server.js",
        "webpack": "node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js"

elhigu:
@pilgylis according to this https://docs.npmjs.com/files/package.json#bin those scripts should be found from `node_modules/.bin` too... or are they some `.cmd` or other scripts that are installed in windows?

pilgylis:
It failed with both node_modules/.bin/webpack and node_modules/.bin/webpack.cmd in my machine

elhigu:
Sounds like direct linking as you've done it would be the most portable way to do it... gotta checkout at home with windows if there is a cleaner way. Im not sure if hardcoding the 'node' name is a good idea either.

pilgylis:
My explanation (mostly guess) is follow. Please correct me if anything is incorrect.

'npm' is in fact the shortcut for 'node npm-cli.js'. Basically when we run 'node something.js', the content of something.js will be put into a js function (with environmental parameters) and run.

In our case, 'webpack' and 'webpack.cmd' are in fact OS batch files, so when put into a js function, it is obviously an error; and 'node webpack.js' works.

Note: I am using Node 7.5

elhigu:
@pilgylis yep, I do understand why `node notnodeprogram.cmd` doesn't work. I'm just thinking if there would be some better way to pass the memory limit parameter than calling node directly from package.json (for example in some debian based machines doesn't have `node` executable but its called `nodejs` so there is a little crossplatform compatibility issue in there too...). 

Btw. I'm not sure if I understood correctly what you meant with the comment how `npm` runs the scripts declared in package.json, but I'm sure it doesn't wrap those scripts to any js function, but just directly executes them in a new process with some additional environment variables given. I haven't checked the code, but I suppose it just calls `spawn` or `exec` to run them.

elhigu:
@pilgylis I updated pull request with your changes and dug out correct node executable path from npm environment variables. Please let me know if it now works for you without any modifications.

pilgylis:
@elhigu doesn't work!

Only the following works: (**node** instead of **npm run node**)
"webpack-dev-server": "**node** --max_old_space_size=4096 node_modules/webpack-dev-server/bin/webpack-dev-server.js",
"webpack": "**node** --max_old_space_size=4096 node_modules/webpack/bin/webpack.js"

elhigu:
@pilgylis thanks for testing, I'll debug it later on when I get my hands on windows box...

elhigu:
Couldn't find any better way than just expect `node` to be found from path.

elhigu:
@vincent-cm looks like you are right, I had different signaling for the AOT build with env variable on my real branch where I copied this. 

But actually the switch allows compiler to use more memory (my build takes ~3GB). I'll update scripts allowing to run AoT build with passing environment parameter.

pilgylis:
@vincent-cm thanks, I didn't notice this.

SherlockStd:
This is definitely a really good PR, need to be merged ASAP !!

PR #1492, opened 2017-02-07
Updated to support font loaders

nikhilsarvaiye:
not sure for which branch i should create a pull request

* **Other information**:

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -179,7 +179,8 @@ module.exports = function (options) {
           exclude: [helpers.root('src/index.html')]
         },
 
-        /* File loader for supporting images, for example, in CSS files.
+        /* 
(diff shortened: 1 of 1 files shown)


config/webpack.common.js
```diff
@@ -179,13 +179,21 @@ module.exports = function (options) {
...
          */
         {
           test: /\.(jpg|png|gif)$/,
           use: 'file-loader'
         },
 
+        /* File loader for supporting fonts, for example, in CSS files.
+        */
+        { 
+          test: /\.(eot|woff|woff2|svg|ttf)([\?]?.*)$/,
```
gilhanan: `woff|woff2` can be optimized to `woff2?`

debslab:
I see that you have replaced url-loader with file-loader. Is there any specific reason for doing so?

PR #1079, opened 2016-10-01
Added watch:test:nocoverage task to fix coverage issues when running …

yjaaidi:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Bug fix.
- **What is the current behavior?** (You can also link to an open issue here)

https://github.com/AngularClass/angular2-webpack-starter/issues/361
- **What is the new behavior (if this is a feature change)?**

Added 'watch:test:nocoverage' task in order to be able to run karma tests with the right source map as the coverage breaks it.
- **Other information**:

…karma in debug by disabling istanbul coverage.

PR #1499, opened 2017-02-08
test: disable coverage in watch mode or when NO_COVERAGE env is present

### LINKED ISSUE #1392: Unit Test Sourcemap Missing
I'm submitting a...
[*] bug report
[ ] feature request

What is the current behavior?

Sourcemapped files aren't being served when debugging unit tests.

1. Run ```npm run watch:test```
2. Open Karma in Chrome
3. Open debugger tools.
4. Files are missing from the sources tab.

shlomiassaf:
NO_COVERAGE is provide for IDE integration

This PR enables proper breaking point in TS code.

Fixes #1392

Approach:
All configurations files (karma/webpack.test) are configured without coverage, if coverage support is needed it will be added into the config's.
This is common sense, there's no point in setting something then removing it which is also not clear and readable.

Coverage is opt-out, i.e. by default you will get coverage reports... the developer needs to opt out of coverage explicitly.

There are 2 ways to opt out:
  - Using `--watch-mode`. watch mode implies development session so no coverage there. This is automatically set as part of the npm scripts in this project.

  - Using `NO_COVERAGE` environment variable, when set the coverage will not run. This is a fallback that is useful for IDE integration where karama is executed and controlled by the IDE.

image

### DIFF
--- a/config/helpers.js
+++ b/config/helpers.js
@@ -22,7 +22,16 @@ function isWebpackDevServer() {
 
 var root = path.join.bind(path, ROOT);
 
+function useCoverage() {
+  if (EVENT.indexOf('--auto-watch') !== -1) {
+    return false;
+  } else {
+    return !process.env.hasOwnProperty('NO_COVERAGE')
+  }
+}
+
 exports.hasProcessFlag = hasProcessFlag;
 exports.hasNpmFlag = hasNpmFlag;
 exports.isWebpackDevServer = isWebpackDevServer;
 exports.root = root;
+exports.useCoverage = useCoverage;
--- a/config/karma.conf.js
+++ b/config/karma.conf.js
@@ -2,6 +2,25 @@
  * @author: @AngularClass
  */
 
+const helpers = require('./helpers');
+
+/**
+ * This object is the diff required to add coverage support to karma.
+ * Every property set here will OVERWRITE the property in the original config, no merges.
+ */
+const COVERAGE_CONFIG_DIFF = {
+  preprocessors: { './config/spec-bundle.js': ['coverage', 'webpack', 'sourcemap'] },
+  reporters: ['mocha', 'coverage', 'remap-coverage'],
+  coverageReporter: {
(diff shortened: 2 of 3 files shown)


shlomiassaf:
After posting this I'v noticed that @yjaaidi has posted PR #1079

It appears to be roughly the same solution....

@yjaaidi my apologies...

I would prefer going with this one since it does not involve adding an extra package.

Please discuss.

gviligvili:
What about ? 
```
....


/**
 * Webpack Constants
 */
const ENV = process.env.ENV = process.env.NODE_ENV = 'test';

const autowatch = process.env.npm_lifecycle_script.indexOf('--auto-watch') !== -1;

....

/**
 * Webpack configuration
 *
 * See: http://webpack.github.io/docs/configuration.html#cli
 */
module.exports = function (options) {
      let config = {
              ...
      };

      // skip coverage in watch mode
      if (!autowatch) {
            config.module.rules.push(/**
                   /**
                   * Instruments JS files with Istanbul for subsequent code coverage reporting.
                   * Instrument only testing sources.
                   *
                   * See: https://github.com/deepsweet/istanbul-instrumenter-loader
                   */
                  {
                        enforce: 'post',
                        test: /\.(js|ts)$/,
                        loader: 'istanbul-instrumenter-loader',
                        include: helpers.root('src'),
                        exclude: [
                              /\.(e2e|spec)\.ts$/,
                              /node_modules/
                        ]
                  }
            );
      }

      return config;
};

```

This is how I disabled the coverage in watch mode,
Sorry if it's not right, just trying help.

shlomiassaf:
@gviligvili 

It's about the same, I just added support for an environment variable for IDEs.

I also removed some karma plugins, even though just removing istanbul does the trick there is still coverage plugins running which has no point...

MangeshPimpalkar:
Thanks it works great.

gdi2290:
@shlomiassaf any updates on this?

PR #1512, opened 2017-02-10
Change the title to angular2-webpack2-starter

DespertaWeb:
* **Other information**:
Please change the title to angular2-***webpack2***-starter, that would tell the webpack version you're running. 

So you don't have to be pocking around in package.json ;)


katallaxie:
🤔 I will close this, because there is everything else in there, but anything that is related to what the description says.

PR #1513, opened 2017-02-10
rename selector 'app' to 'app-starter'

dherges:

rename selector 'app' to 'app-starter'

'app'

'app-starter'

* **Other information**:

see angular style guide: https://angular.io/styleguide#!#02-07

also, solves html validation error: "Element app not allowed as child of element div in this context. (Suppressing further errors from this subtree.)", see https://validator.w3.org/nu/

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -13,7 +13,7 @@ import { AppState } from './app.service';
  * Top Level Component
  */
 @Component({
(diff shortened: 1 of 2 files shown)


katallaxie:
🎱 No, this is not the starter app, but a starter for your next project. So, let's keep it. I will close the PR.

Issue #1468, opened 2017-02-03
Change config folder structure and npm task related


michaeljota:
* **I'm submitting a ...**
- [ ] bug report
- [x] feature request
- [x] question about the decisions made in the repository

* **What is the current behavior?**

The config folder has config files for every component in it. Also, some task in npm scripts are repetitive, and can be handle in a better way using webpack env argument. 

* **What is the wanted/proposed behavior?**

I fork this project to add some custom things, but in the way I manage to structure in a more scalable way (according to me). And I would like to propose that change in for the config folder in this project. I would made a pull request, but I want to know if this is something you would like. 

`config structure`

![captura de pantalla 72](https://cloud.githubusercontent.com/assets/10507776/22598199/045561dc-ea09-11e6-8cba-aea44e6997e3.png)

The webpack folders have an index that resolve and return the proper configuration.

* **What is the motivation / use case for changing the behavior?**

Adding additional build scenarios aren't intuitive and it's hard to think how exactly do it.

* **Please tell us about your environment:**

- Angular version: 2.4.x
- Browser: [all]
- Node: 6.9.4
- npm: 3.10.10

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

JoniJnm:
I think some refactor should be done too:

# Rename webpack.*.ts files

Webpack should have 3 config files:

1. webpack.dev.js: development mode: no minimize files, create sourcemap...)
2. webpack.test.js: for unit testing
3. webpack.build.js: build aot or jit

The file webpack.common.js should be renamed too because it's not used in webpack.dev.js. It could be webpack.helpes.js or webpack.utils.js and it should have functions that could be used in the other config files to inject some plugin, config or whatever.

# Create real eviroment config files

The project should have different enviroment config files:

* config.dev.js
* config.pre.js
* config.pro.js
* config.github.js

In these files the user could set the API URL, the domain used, the color of the topbar or what they want.

DefinePlugin should add these config to use in .ts files. Something like:

```
//save config from process.env in a dictionary (function in webpack.utils.js)

const keys = ['ENV', 'API_URL', 'PUBLIC_KEY', 'USE_SOME_FEATURE'];
let config = {};
for (var i=0; i<keys.length; i++) {
	let key = keys[i]
	try {
		config[key] = JSON.parse(process.env[key]);
	}
	catch(e) {
		config[key] = process.env[key];
[...]

Issue #1368, opened 2017-01-12
Discussion: Drop Node 5 Support?


colinskow:
Protractor `^5.0.0` dropped support Node 5. If we want to upgrade Protractor further we will need to drop support for Node 5. Other deps will likely do the same in order to use ES6 code natively.

In my opinion we should focus on Node versions from LTS to latest.

d3viant0ne:
Touchy subject as many people are for a reason I honestly can't understand still using node 4.x.

I personally agree, support LTS + Next with a grace period for the previous LTS & dev version. If we are going to do this, we need to put out a warning via pretty much every way possible and give it a few weeks to soak in before we pull the trigger imo.

shlomiassaf:
ohhh no, winter is coming :)

I would go LTS... 

This is a starter project, it should use the latest and greatest with responsibility... so LTS sound good.

katallaxie:
👍

d3viant0ne:
To be clear, I completely agree with the current LTS + Next approach we just need to be ready for the torches & pitch forks from the slow adopters when we drop 4.x support :)

colinskow:
Angular 2 (and thus we) dropped support for 4.x some time ago. Node 5 passed quickly, so I don't expect many people stuck with it.

The reason 6 is important is because it has full ES6 support. That's why things are breaking on previous versions. I expect future versions should be stable for a while until Node starts adding ES 2016/17 support perhaps in 8.

shlomiassaf:
Just a side note, we can still run < LTS with some scripting before running...
Something like running removing protractor 5 installing 4 and just running the build & tests w/o e2e...

This makes < LTS support limited but there...

JonnyBGod:
I agree with the LTS + Next version. Node already defines road map for versions.

A simple version pairing note on readme should do the trick for the less informed.

PR #1515, opened 2017-02-10
Webpack config folders alternative structure

michaeljota:

Feature about the Webpack configuration

The Webpack configuration it's called individually for all the run. This is somehow explicit about the configuration, but it is not good for extends. 

This PR should introduce a new structure for the Webpack configuration folders. This should be easy to understand, and to extend. The webpack should be called without any config argument, and the main Webpack config file should call the Webpack config folder. This `index.js` should be able to resolve the request configuration and return it. 

* **Other information**:

  - It should work now for common browsers build. 
  - Github deploy still WIP. 
  - This code needs Node 6 to run (Destructive imports, parameters initialization). But as it's being discussed here (#1368) Node 5 it's not supported by Angular 2, so, effectively this doesn't support anything below Node 6 LTS either. 

Ref: #1468 
This is other approach from #1508. This works. You can test it to merge.


dereklin:
Is it true I still need to repeat things like this in multiple config files?
```
      new DefinePlugin({
        'HMR': METADATA.HMR,
        'process.env': {
          'NODE_ENV': JSON.stringify(METADATA.ENV),
          'HMR': METADATA.HMR,
        }
      }),

```

If the plugins are defined in common, I want to be able to inherit them from common and be able to extend them.  The same goes for other settings.

As for now, I find myself having to copy paste the same config file and modify them a little bit here and there.  It's hard to maintain when there are many different environment settings and different developer settings.

Maybe one file will work better?: 
https://dev.to/alflennik/the-fine-art-of-the-webpack-2-config

gdi2290:
sorry this PR changed a lot and I didn't get to it. I'm going to refactoring the repo so the folder structure will change

PR #1528, opened 2017-02-14
refactor(package.json): run clean task at the time of build

shamsher31:
Run clean task at the time for build both for dev and production.

While running `build` task, it does not remove `compiled` directory which consist of Angular`s ngFactory files. Which through compile error even while building for development environment.
It is always good to freshly install dependencies on production to avoid such issues.

error

It will freshly install all the dependencies and create `dll` and `compile` files. Which avoids unexpected error during build time.

* **Other information**:
Now when you run `npm run build` or `npm run build:aot`

1. Clean task will run which will remove `node_modules doc coverage dist compiled dll`.
2. And `npm install` will run which will install all the dependencies and pass webpack config for build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,9 +11,9 @@
   "homepage": "https://github.com/angularclass/angular2-webpack-starter",
   "license": "MIT",
   "scripts": {
-    "build:aot:prod": "npm run clean:dist && npm run clean:aot && webpack --config config/webpack.prod.js  --progress --profile --bail",
+    "build:aot:prod": "npm run clean && npm install && webpack --config config/webpack.prod.js  --progress --profile --bail",
     "build:aot": "npm run build:aot:prod",
-    "build:dev": "npm run clean:dist && webpack --config config/webpack.dev.js --progress --profile",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,9 +11,9 @@
   "homepage": "https://github.com/angularclass/angular2-webpack-starter",
   "license": "MIT",
   "scripts": {
-    "build:aot:prod": "npm run clean:dist && npm run clean:aot && webpack --config config/webpack.prod.js  --progress --profile --bail",
+    "build:aot:prod": "npm run clean && npm install && webpack --config config/webpack.prod.js  --progress --profile --bail",
```
manolakis: this change causes friction with [yarn](https://yarnpkg.com) users, isn't it?

shamsher31: We always have flexibility to modify this as per our project requirements. Currently I don't see any yarn config in this repo. Anyway we can ask for suggestions.

gdi2290:
I'm going to redo the build system later

PR #1576, opened 2017-02-27
add vars from external config files

elvis-macak:
Feature
Help it easier to add new variables in code and different environments, make use of the webpack DefinePlugin

Consider my case: 
When I want to add a new config, like `Google Analytics token`, they should be different in development/test/production mode. (I think it would be better to add it into DefinePlugin)
So first I add `GA_TOKEN` to each `webpack.*.js`'s DefinePlugin: 
```
      new DefinePlugin({
        'HMR': METADATA.HMR,
+        'GA_TOKEN': 'this is a setting in dev',
        'process.env': {
          'NODE_ENV': JSON.stringify(METADATA.ENV),
          'HMR': METADATA.HMR,
        }
      }),
```

And then add it into `custom-typings.d.ts`:
```
declare var GA_TOKEN: string;
```

Finally, I can use the GA_TOKEN in my ts code, like the ENV variable which already defined and declared.

When my settings getting growth, like `PAYMENT_TOKEN`, `BACKEND_SERVER`, thus coding style gets explosion.

Just simply define a GA_TOKEN in the config/vars/development.js and config/vars/prod.js, and easily use `VARS.GA_TOKEN` in your ts code.
No other configs need to be changed.

* **Other information**:

While export all the config out into a single file, it would make it easier to be encrypt and generated

### DIFF
--- a/config/vars/common.js
+++ b/config/vars/common.js
@@ -1,0 +1,4 @@
+module.exports = {
+  GA_CODE: 'UA-91111111-1',
+  HOME_TOKEN: '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI',
+}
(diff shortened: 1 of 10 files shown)


gdi2290:
interesting. I might end up going down this route

gdi2290:
I'm going to redo this and most likely go with something similar to what you did

PR #1581, opened 2017-02-28
 feat(webpack): ignoring mock-data dir when prod

gilhanan:
feature

mock-data files being copy in prod build

mock-data files not being copy in prod build

I discovered it's very useful to use mock-data files both in development and test, but it's bit annoying when publishing to prod, these files usually unnecessary.


gdi2290:
thanks!

Issue #1703, opened 2017-04-27
Webpack cache busting not fully implemented


DBosley:
* **I'm submitting a ...**
[ ] bug report
[X] feature request
[ ] question about the decisions made in the repository

* **What is the current behavior?**
If you're using lazy loading with angular modules these all generate chunk files in the webpack output. Webpack creates a manifest of chunk references to file names. By default, this manifest is loaded into the first defined webpack entry point's output file. When the manifest changes, however, the chunkhash for that specific output file isn't guaranteed to change. In the current project, 'Polyfills' is the first entry point. This file very rarely changes and the chunkhash for it will always be the same. This means the manifest never gets updated and will contain old file references when loaded from cache.

* **What is the expected behavior?**
the webpack manifest should be extracted from the output and injected into the index html

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

Webpack documentation on how to implement proper chunkhash caching:
https://webpack.js.org/guides/caching/

jkreileder:
Like this? https://github.com/jkreileder/angular-lightning-webpack-starter/commit/91ec5b81b64ec58c04094d1ac5cf82a823d320d5
Will send a pull request if there's interest.

colinskow:
It looks like our output generates MD5 hashes that are embedded in the names of the source files. I thought this was the primary mechanism to ensure correct caching.

This sounds like a valuable pull request, but I would like to better understand the problem we are trying to solve.

jkreileder:
Yes, but you want to keep cache busting as minimal as possible. Currently each change to the application code also invalidates the polyfills bundle:

Original:
```
0.420d6c76fa8938a2547f.chunk.js
1.cc39a588cf9a6aab422d.chunk.js
2.cd20cdf22c5651d965f6.chunk.js
5.4c102ffef891d7aede1c.chunk.js
6.cea5ea445288c79e42e4.chunk.js
main.0a01f4183ca940373589.bundle.js
polyfills.8258808d5bf4b9758918.bundle.js
vendor.a2ff3b67548bea4e88b8.bundle.js
```
Now change, for example, 'Hello from Barrel' to 'Hi from Barrel' in src/app/+barrel/barrel.component.ts
Result:
```
0.420d6c76fa8938a2547f.chunk.js
1.da087417ad3cced709d8.chunk.js
2.cd20cdf22c5651d965f6.chunk.js
[...]

PR #1712, opened 2017-05-02
Improve cache (busting) configuration

jkreileder:
This PR improves cache busting by keeping module IDs stable, extracting webpack's manifest to its own file and inlining that file in index.html

Without these changes the manifest was included in the polyfills bundle. Which means changes to the main app also invalidated caching of the polyfills bundle. Also, the module IDs were not stable up to now.

Cf.
- #1703 
- https://survivejs.com/webpack/optimizing/separating-manifest/#extracting-a-manifest
https://survivejs.com/webpack/optimizing/adding-hashes-to-filenames/#enabling-hashedmoduleidsplugin-
- https://webpack.js.org/guides/code-splitting-libraries/#manifest-file
- https://webpack.js.org/guides/caching/

### DIFF
--- a/config/webpack.prod.js
+++ b/config/webpack.prod.js
@@ -75,7 +75,7 @@ module.exports = function (env) {
        *
        * See: http://webpack.github.io/docs/configuration.html#output-sourcemapfilename
        */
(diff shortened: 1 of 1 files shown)


gdi2290:
thanks!

PR #1714, opened 2017-05-02
Added 'IS_PROD' flag

### LINKED ISSUE #1700: Check DEBUG mode in TypeScript code (similar to checking isDevServer flag in index.html)
## Question
**How can I check in the TypeScript code that I'm debugging the app and it is NOT running in the production mode?**

The `webpack.common.js` file has `isDevServer` flag, which indicates running local dev server and can be used in HTML files (it's used in the default `index.html`). Is there an option to use this flag in the TypeScript code (Angular classes)? Or maybe there is something similar.

## Longer rant
A very simple app has no backend and is using Azure AppInsight for logging. I don't want to pollute the log with messages generated during debugging the app locally (those records could be filtered out, but I'm better off preventing them).
Of course, there are alternative options:

- One would be a URL check (if it's http://localhost, then not to log), but in some cases it could be http://0.0.0.0. 
 - Then I thought to check the port number, but it will become vulnerable to future port changes (happened once in the past), as such dependency on the port would be not obvious.

Maybe there are other options, but I see handling the "_debug_" mode as the most intuitive for future developers.

Last addition, the logging is done in a service, which gets injected. Hence, maybe a good practice would be to wire up a different service in the debug mode, but again, what's the best way to do that.

AKlaus:
Added 'IS_PROD' flag to indicate whether the running environment is production or not. Resolves #1700.

Now it's easier to run 'debug-only' or 'production-only' functionality in the TypeScript code just by checking this flag.

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -146,6 +146,7 @@ module.exports = function (options) {
       new DefinePlugin({
         'ENV': JSON.stringify(METADATA.ENV),
         'HMR': METADATA.HMR,
+		'IS_PROD': false,
         'process.env': {
           'ENV': JSON.stringify(METADATA.ENV),
           'NODE_ENV': JSON.stringify(METADATA.ENV),
--- a/config/webpack.prod.js
+++ b/config/webpack.prod.js
@@ -157,6 +157,7 @@ module.exports = function (env) {
       new DefinePlugin({
         'ENV': JSON.stringify(METADATA.ENV),
         'HMR': METADATA.HMR,
+		'IS_PROD': true,
         'process.env': {
           'ENV': JSON.stringify(METADATA.ENV),
           'NODE_ENV': JSON.stringify(METADATA.ENV),
--- a/config/webpack.test.js
+++ b/config/webpack.test.js
@@ -197,6 +197,7 @@ module.exports = function (options) {
       new DefinePlugin({
         'ENV': JSON.stringify(ENV),
         'HMR': false,
+		'IS_PROD': false,
         'process.env': {
           'ENV': JSON.stringify(ENV),
           'NODE_ENV': JSON.stringify(ENV),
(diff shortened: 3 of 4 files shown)


SherlockStd:
really nice, was searching for this yesterday 😂

bartholomej:
Or you can use it like this:
```js
if (process.env.ENV === 'production') {
  console.log('Do some magic in prod build');
}
```

AKlaus:
@bartholomej, IMHO, relying on hard-coded strings (the '_production_' in this case) isn't a good practice. Therefore, a variable would come in handy (the name of the variable can be even more verbose, e.g. `IS_PRODUCTION`).

config/webpack.dev.js
```diff
@@ -146,6 +146,7 @@ module.exports = function (options) {
       new DefinePlugin({
         'ENV': JSON.stringify(METADATA.ENV),
         'HMR': METADATA.HMR,
+		'IS_PROD': false,
```
thinkingmedia: This has to be JSON.stringify(false). Please read the documentation for DefinePlugin.

config/webpack.prod.js
```diff
@@ -157,6 +157,7 @@ module.exports = function (env) {
       new DefinePlugin({
         'ENV': JSON.stringify(METADATA.ENV),
         'HMR': METADATA.HMR,
+		'IS_PROD': true,
```
thinkingmedia: JSON.stringify(true)

config/webpack.test.js
```diff
@@ -197,6 +197,7 @@ module.exports = function (options) {
       new DefinePlugin({
         'ENV': JSON.stringify(ENV),
         'HMR': false,
+		'IS_PROD': false,
```
thinkingmedia: JSON.stringify(false)

thinkingmedia:
This functionality already exists. As @bartholomej has pointed out. `process.env.ENV` is less likely to collide with other variables than `IS_PROD`. Understand that the `DefinePlugin` performs a search and replace. So longer more unique identifiers are preferred.

PR #1718, opened 2017-05-04
perf(dev): enable useCache and disable tslint-loader

admosity:
Enable `useCache` on `awesome-typescript-loader` and remove `tslint-loader` from dev config.

tslint runs over all the files again on every run and this is slow on a large application. With the `useCache` option, the initial startup time I saw was improved from 20s to 10s on a large application. Here is the result from the base starter though:

*Before*
image

*After*
image

* **Other information**:
I think that since linting is part of the testing pipeline, it can be done there. I've seen that the tslint --fix option has been fairly powerful as well. With the power of new in editor linting options, I think it would be better to disable this for rebuild performance.


admosity:
New before and after on a much larger app with a warmed up awcache:

### Before
image

### After
image

gdi2290:
thanks!

sentika:
Not a good solution, because while tslint-loader is disable you can't see lint issues in dev mode.
Manual running lint it's not a good idea too

PR #1731, opened 2017-05-10
chore(Dockerfile): Use nginx:alpine for a smaller image size

cmckni3:


gdi2290:
can you explain more of the difference here other than just the title

thinkingmedia:
Alpine is a much smaller Linux image (5mb). I'm not sure this should be the default. It would be better to just add a comment in the Dockerfile to notify people that there is a smaller version. For production, people prefer stability over memory limits.

cmckni3:
Yes, alpine makes the built image much smaller and it builds much faster.

I think alpine is pretty stable. The ability to easily install new software on top of the image is the only advantage I see with using the debian image.

gdi2290:
as @thinkingmedia mentioned I rather it be a comment for users to opt into

cmckni3:
ok see #1732

Issue #1498, opened 2017-02-08
JS Heap out of memory when running build:aot:prod


doorman02:
**Note: for support questions, please use one of these channels:** [Chat: AngularClass.slack](http://angularclass.com/member-join/) or [Twitter: @AngularClass](https://twitter.com/AngularClass)

* **I'm submitting a ...**
[X ] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Do you want to request a *feature* or report a *bug*?**
Reporting a bug

* **What is the current behavior?**
When I run npm run build:aot:prod

I get the following error:
==== JS stack trace =========================================
Security context: 00000133277CFB61 <JS Object>
    1: enter [C:\MyApp\node_modules\optimize-js\lib\index.js:~13] [pc=00000269B6CE8F25] (this=000002BBF4E471E1 <an Object with map 000000D02DAA96D1>,node=0000022D23C7E1B9 <a Node with map 000003DD7E69E889>,parent=0000023AC9B9F6D9 <a Node with map 000003DD7E694801>)
    2: arguments adaptor frame: 4->2
    3: visit(aka visit) [C:\MyApp\node_m...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:AvJOMERrnz94ekVua0u5).

* **What is the expected behavior?**
To build successfully

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**

- Angular version: 2.4.6

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

pilgylis:
I am having this issue too. Maybe it is a leakage somewhere. 

On the other hand, is there any way to increase the maximum RAM? I observed it crashed at around 1.5GB.

shlomiassaf:
I don't think this is an issue with the starter.
Can you open it in the [optimize-js](https://github.com/nolanlawson/optimize-js)

shlomiassaf:
We also need more information on how to reproduce it.

If I just clone the repo it doesn't happen...

pilgylis:
[...]

PR #1734, opened 2017-05-11
Update package.json

nickppa:
Add cross-env to setting the env across the different platforms.

Bug fix

When I use "npm run build:aot" in windows 8, it will meet an error, it said the "BUILD_AOT" is not a command, so I added this.

* **Other information**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "homepage": "https://github.com/angularclass/angular2-webpack-starter",
   "license": "MIT",
   "scripts": {
-    "build:aot:prod": "npm run clean:dist && npm run clean:aot && BUILD_AOT=1 npm run webpack -- --config config/webpack.prod.js  --progress --profile --bail",
+    "build:aot:prod": "npm run clean:dist && npm run clean:aot && cross-env BUILD_AOT=1 npm run webpack -- --config config/webpack.prod.js  --progress --profile --bail",
     "build:aot": "npm run build:aot:prod",
     "build:dev": "npm run clean:dist && npm run webpack -- --config config/webpack.dev.js --progress --profile",
     "build:docker": "npm run build:prod && docker build -t angular2-webpack-start:latest .",
@@ -104,6 +104,7 @@
     "awesome-typescript-loader": "~3.1.2",
     "codelyzer": "~2.1.1",
     "copy-webpack-plugin": "^4.0.1",
+    "cross-env": "^4.0.0",
     "css-loader": "^0.28.0",
     "exports-loader": "^0.6.4",
     "expose-loader": "^0.7.3",


colinskow:
I suggest we remove `BUILD_AOT=1` to keep it clean. I there are reason this was changed?

SherlockStd:
@colinskow This env variable was introduced by @dereklin on https://github.com/AngularClass/angular-starter/issues/1498#issuecomment-287596548 because the AOT compilation wasn't working. I added it to my project and this is efficient, so I suggest not to remove it.

colinskow:
@SherlockStd it seems to me like #1498 was an issue with Angular 4 beta. It also doesn't make sense how an environment variable (not even used by the compiler) would fix an issue. I think it was fixed by updating Angular.

I understand an environment variable can be useful in build processes running outside NPM scripts, but is it necessary inside the NPM script too?

Otherwise if we're going to keep the environment variable in the script this PR needs to be be merged.

SherlockStd:
@colinskow It was fixed by adding `process.env.BUILD_AOT` into `webpack.common.js` here: https://github.com/AngularClass/angular-starter/commit/6fb1a2b7d8af3a5e7983987c761f94ea310a4bad#diff-f737fedb5357468044db2dbe98eb0218R29

SherlockStd:
Last version of Angular 4 didn't fix this

colinskow:
I'm not opposed to merging this if the fix really works. There's just no logical explanation why an environment variable would fix a bug in @angular/compiler that causes it to run out of memory. I searched the source code of Angular, `ngc-webpack` etc. and there are no mentions of AOT_BUILD.

@SherlockStd  can you confirm that the bug appears as soon as you remove the ENV variable from the script?

We should find the root cause rather than doing a bandaid patch like this.

nickppa:
@colinskow pls see the the commit "496cf6b9872b01c2c30b33356dad167e13c02dea", you are using
 `"build:aot:prod": "npm run clean:dist && npm run clean:aot && npm run webpack -- --config config/webpack.prod.js  --progress --profile --bail" `
**instead of** 
`"build:aot:prod": "npm run clean:dist && npm run clean:aot && webpack --config config/webpack.prod.js  --progress --profile --bail"`

And in the helpers.js, you are using `process.env.npm_lifecycle_event` to get the command, for example, if you are running `"npm run build:aot:prod"`, then you will get `"build:aot:prod"` for `process.env.npm_lifecycle_event`.

But you will not get `"build:aot:prod"` after this commit, you will get `"webpack"` instead, because you are using `"npm run webpack -- --config config/webpack.prod.js  --progress --profile --bail"` in the script, I guess you can't use the script `"webpack": "node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js"` which is defined in your package.json when you are using `"webpack --config config/webpack.prod.js  --progress --profile --bail"` not `"npm run webpack -- --config config/webpack.prod.js  --progress --profile --bail"`, you fixed the "**out of memory**" bug use the parameter `"--max_old_space_size=4096"` and `"npm run"` to use this script in `"build:aot:prod"`.

But you brought another bug in, the `process.env.npm_lifecycle_event` would not get the right string, it would get `"webpack"` instead. So when you are using `"const AOT = helpers.hasNpmFlag('aot');"` in webpack.common.js, you will get the `AOT = false`, then you will not enable the aot.

So you are using the `process.env.BUILD_AOT`.

SherlockStd:
@colinskow no `BUILD_AOT` is not fixing the `out of memory bug`, it's fixing the AOT build not launching with the npm tag (as explained very well by @nickppa)

colinskow:
@nickppa and @SherlockStd thank you very much for the explanation. This makes sense to me now. It would be nice to find a less hacky solution but I suggest we merge this so it works on Windows.

CC: @gdi2290

gdi2290:
removing BUILD_AOT would be preferred

admosity:
This PR is defunct now since #1744 is merged.

PR #1736, opened 2017-05-11
FEATURE: Support for .env.js file with local environment dependent variables

kamil-kielczewski:

Feature: Add support for .env files which contains local environment variables (different for dev, test, production environments) like e.g. google_analytics_tracking_id or sentry_io_DSN, google_api_key etc. This is very useful feature for instance for **docker image build**

This approach (usage .env files) is popular in backend frameworks e.g. [PHP Laravel](https://laravel.com/docs/5.4/configuration#environment-configuration). The main idea is to put all local environment dependent variables into single file to easy manage them.

Because index.html file is generated by webpack before typescript is set up the .env file must be written in pure js (so the name is .env.js) - but because body of this file is very simple it is no problem. 

Last thing: in repository exist only .env.example.js file and during instalation this file should be copy as .env.js file (which is local dependent so excluded from repository) and edit.

Curently there is no single simple config file where we can put all local environment dependent variables to manage them is not easy especially when the project is large and has many that variables.

To compile project .env.js file must exist (by copy .env.example.js). As default google_analytics_track_id variable inside this file is NULL so in index.html file section with GA script will not appear. In Home component I introduce (as example) how to get access to variables from .env.js file.

* **Other information**:

I tested it for dev, prod and AoT build on MacOs


kamil-kielczewski:
https://laravel.com/docs/5.4/configuration#environment-configuration

kamil-kielczewski:
I also want to say that in **.env.js** file you can also keep **API main url** (different for production, dev, testing) and urls for other envirnment dependent services used by app.

kamil-kielczewski:
Hi @SherlockStd @colinskow @nickppa can I know your opinion about this feature? I will gratefull for every comment.

SherlockStd:
Really nice ! Waiting for the merge ! :)

kamil-kielczewski:
Hey @gdi2290 , I upgrade a little my solution:

1. At first running `npm start/build` if `.env.js` not exist then is created from `.env.example.js` (in `webpack.common.js`) - so project always run - (in previous version build fail if user not create by chand `.env.js` file). So project is compliable after git clone. 

2. I remove non portable 'cp' command from README.md - now it is no needed any more.

3. I add 'API_URL' to `.env.example.js` to suggest user where he should put this kind of environment dependent variable.

SherlockStd:
I need you to answer 3 questions for more clarity:
- So is this `env` now available on Angular ?
- How can we access it ?
- How can we modify it's content after prod compilation ?

Thanks in advance

kamil-kielczewski:
1. I don't know that I understand the question - currently this is available only on this pull request :P

2. In `src/app/home/home.component.ts and html `is example how to get access to env inside controllers. And also in `src/index.html` is example (on google_analytics) how to get access in "webpack html templates"

3. The idea is NOT to modify value of this variables AFTER compilation. The idea is to set proper values to each environment BEFORE compliation - and `.env.js ` will be compiled into `dist`. If you want to change some environment values you should rebuild application. But if you wanna something more dynamicaly changed you should use other approach (like for example may be read variables from dedicated backend API)

SherlockStd:
1. I was talking about our ANgular app on this repo, but the questions 1 & 2 was stupid, I removed them :smile: 
3. It will be nice to be able to, because suppose you want to ship a compiled bundle, it's impossible to reuse it because we can't change env vars

kamil-kielczewski:
ad 2 Maybe in future I will think about this kind of solution where .env.js file is copy into 'dist' and main.bundle.js file read it. I created current solution because i needed this 'before compilation' .env.js file in many projects that I create using docker. So current solution is also helpful but of course in future we can improve it.

kamil-kielczewski:
Hi @gdi2290 I would like to add that this pull request is in fact continuation of [this one(dockerfile)](https://github.com/AngularClass/angular-starter/pull/1772). This feature will allow following flow for deployment process:

```
git clone/pull ...
cp /path/to/env/file/.env.js ./.env.js
docker build -t angular-starter .
docker run --name angular-starter ...
```

Where in [.env.js](https://github.com/kamil-kielczewski/angular-starter/blob/5641ba40b274990d81b165897fcbfe2dcddabdcb/.env.example.js) file which contains all environment dependent variables like backend_api_url etc.

This is very handy and allow to automate deployment process. Of course may be exist better solution to achieve this goal but for this moment this solution is enough.

kamil-kielczewski:
@gdi2290 Storing configuration in the environment separate from code is based on [The Twelve-Factor](https://12factor.net/config) App methodology

kamil-kielczewski:
Reading this and also [this](https://www.jvandemo.com/how-to-configure-your-angularjs-application-using-environment-variables/) I realized that to achieve goal of .env which can be change without recompiling application is to remove webpack index.html generation part because this not allow to achieve this goal.  So I plan to work on that approach and in fact this pull request can be closed @gdi2290 - but please give me somme comment do you are open to support .env.js files which can be changed without app recompilation?

gdi2290:
sorry I didn't get to this in time. I'll be adding better env support with something similar

PR #1742, opened 2017-05-15
[TEMP] minimize config for dll bundle

abelousov:
create minimal config that still supports Dll bundle


abelousov:
sry, opened by mistake

PR #1744, opened 2017-05-16
BUILD_AOT Environment fix for windows builds

tsukhu:
Bug Fix

The windows build for npm run build:aot fails on windows as the environment variable does not get set as we get this error

'BUILD_AOT=1' is not recognized as an internal or external command,
operable program or batch file.

This fix will work on both Windows and Unix/Linux environments as cross-env takes care of this.

* **Other information**:


SherlockStd:
https://github.com/AngularClass/angular-starter/pull/1744/commits/dddc537f1e5b35ae5a3f020b2e383dbbe1ecafe5
You need to link to `node_modules/.bin/cross-env` or install globally. Please fix

tsukhu:
@SherlockStd .. not sure why you need it globally . We have it in the dev dependencies already added in the package.json
"cross-env": "^5.0.0"

gdi2290:
is this good?

SherlockStd:
@tsukhu if you call it just with `cross-env`, not regarding on the OS, it will look at the path, and since the `node_modules` isn't in your path, I don't think it can work if it isn't installed globally, unless you link to the executable as I suggested, just like it's done for Webpack binaries:
```javascript
"webpack-dev-server": "node --max_old_space_size=4096 node_modules/webpack-dev-server/bin/webpack-dev-server.js",
"webpack": "node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js"
```

admosity:
@SherlockStd The `scripts` field is adds `.bin` to your path with npm:
https://docs.npmjs.com/misc/scripts#path

From the CI build, it's running
image

SherlockStd:
Oh :astonished: My bad, thanks for the info :smile:

tsukhu:
We are all good then!!

tsukhu:
Added the lint fix as well now .. There was a build failure due to other updates related to lint

PR #1772, opened 2017-05-31
FEATURE: Dockerfile which totally automate deployment process

kamil-kielczewski:

FEATURE: One of the goal of docker is ability to run project on any machine without **dependency hell**. So to run project we should need only isntalled **OS**, **git** and **docker** (and that's all - no node.js etc.). The proposed solution meets these expectations. We are able to completely automate deployment process (portable on all platforms which support  docker).

In README.md file are included more informations (and instructions).

Current docker file is write in not good way because we need install many dependencies in our host operating system before we are able to run project (like node.js). In fact, if we clone repo and try to build docker image and run it - we fail. 

New dockerfile totally automate build and run application proces. This file can be run in two scenarios (details in README.md): 

1. fast - to build big image which can be fast rebuild (good for staging/testing env)
2. small - to build small image (~20MB) but rebuild process take same time as first build (slow)

In README.md file is also included solution of how to run image in sub-domain (virtual host)

* **Other information**:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,39 @@
+.git
+.github
+.vscode
+coverage
+
+# OS generated files #
+.DS_Store
+ehthumbs.db
+Icon?
+Thumbs.db
+
+# Node Files #
+node_modules
+npm-debug.log
+npm-debug.log.*
+
+# Typing #
+src/typings/tsd
+typings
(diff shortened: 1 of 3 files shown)


gdi2290:
LGTM

PR #1799, opened 2017-06-13
refactor(index.html): async GA, baseUrl, comments

gdi2290:


src/index.html
```diff
@@ -1,28 +1,38 @@
 <!DOCTYPE html>
 <html>
 <head>
+  <title><%= htmlWebpackPlugin.options.title %></title>
```
jiverson: Pretty sure this should be below the `meta` since some browsers only look at the first 1024 bytes before choosing an encoding.

gdi2290: kool, updated

PR #1836, opened 2017-07-23
chore(config): remove json-loader as no longer needed

DanielKucal:
It removes `json-loader` according to [official webpack guide](https://webpack.js.org/guides/migrating/#json-loader-is-not-required-anymore).

For `json-loader@0.5.6` it was throwing the following error several times in CLI and browser console:
image
Although was fixed in 0.5.7

Behavior remains exactly the same, but the config is slighter.

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -147,16 +147,6 @@ module.exports = function (options) {
         },
 
         /**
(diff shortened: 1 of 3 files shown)


gdi2290:
thanks

PR #1839, opened 2017-07-24
Update to Angular v4.3

DanielKucal:
It updates Angular version to ~4.3. 
There were not any breaking changes according to [Angular changelog](https://github.com/angular/angular/blob/master/CHANGELOG.md) and we are using v4.3 from some time on a quite big project without any issues.

Currently it uses v~4.1


gdi2290:
thanks

PR #1848, opened 2017-08-02
add PUBLIC env var for webpack dev server's public property

patricknazar:

Allow's overriding webpack dev server public property. I use it in my case where in eclipse che it assigns an ephemeral port, and my dev machine is another one on the network, 10.0.0.1 (or dev in /etc/hosts) not 127.0.0.1. I can then inject with a macro ie `PUBLIC=dev:${server.3000.port} npm run start`

It's always port 3000 and HOST doesn't give us hostname choice, just bind ip

Defaults to undefined and causes no change. It allows above.

* **Other information**:


gdi2290:
LGTM

PR #1849, opened 2017-08-03
chore: upgrade ngc-webpack to 3.2.0 and remove resourceOverride

shlomiassaf:
Now when with the AOT cleanup loader there is no need to use resourceOverride and it is not recommened for use as it produced some
side effects with assets.


gdi2290:
thanks!

PR #1863, opened 2017-08-15
change env var from PUBLIC to PUBLIC_DEV #1859

### LINKED ISSUE #1859: Not suitable to use `process.env.PUBLIC` on Win
* **I'm submitting a ...**
[x] bug report
[ ] feature request

* **Do you want to request a *feature* or report a *bug*?**
    * Reporting a *bug*

1. Windows 10
1. clone master that is after [e1f25d9] (commited by @patricknazar )
1. `npm start` 

    * My browser try to open `c/Users/Public/` when exec `npm start`.
    * This comes from
        ```javascript
        const PUBLIC = process.env.PUBLIC || undefined;
        ```
        in `/config/webpack.dev.js`
        * https://github.com/AngularClass/angular-starter/blob/master/config/webpack.dev.js#L24

* **What is the expected behavior?**
    * Browser try to open `http://${HOST}:${PORT}`.

* **What is the motivation / use case for changing the behavior?**
    - Windows set the environment variable `PUBLIC=c:\Users\Public`
    - But most of windows users don't notice `process.env.PUBLIC` has already been set.
    - IMO, it's better to define the different env var instead of `PUBLIC`
        - e.g.
        ```javascript
        const PUBLIC = process.env.PUBLIC_DEV || undefined;
        ```

* **Please tell us about your environment:**
    - Angular version: ~4.3.1
    - angular-starter: [e1f25d9]
    - Browser: Firefox 52.3.0 (32 bit)

patricknazar:

fixes #1859 

uses PUBLIC env var which conflicts with windows

uses PUBLIC_DEV instead

* **Other information**:

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -21,7 +21,7 @@ const LoaderOptionsPlugin = require('webpack/lib/LoaderOptionsPlugin');
 const ENV = process.env.ENV = process.env.NODE_ENV = 'development';
 const HOST = process.env.HOST || 'localhost';
 const PORT = process.env.PORT || 3000;
-const PUBLIC = process.env.PUBLIC || undefined;
+const PUBLIC = process.env.PUBLIC_DEV || undefined;
 const HMR = helpers.hasProcessFlag('hot');
 const METADATA = webpackMerge(commonConfig({env: ENV}).metadata, {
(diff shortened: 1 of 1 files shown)


gdi2290:
thanks

PR #1896, opened 2017-10-03
FEATURE: support for .env files (loaded at app bootstrap in browser)

kamil-kielczewski:

Feature: Add support for .env files which contains local environment variables (different for dev, test, production environments) like e.g. backend_api_url, debug_mode, google_analytics_tracking_id or sentry_io_DSN, google_api_key etc. This is very useful feature for instance for **docker image build**, and also if someone deploy app by sending **dist** folder to client, then client only change **dist/.env** file (not need to recompile code). 

This approach (usage .env files) is popular in backend frameworks e.g. [PHP Laravel](https://laravel.com/docs/5.4/configuration#environment-configuration). The main idea is to put all local environment dependent variables into single file to easy manage them. This approach is recommended by [The Twelve-Factor App Methodology](https://12factor.net/config)

This is improved version of this #1736 (and is connected with #1772) (here .env file is loaded in bowser, instead of loading on webpack compilation step)

No support for .env files

After compilation process, .env file will be copied into dist and dynamically loaded on app bootstrap. For instance google_analytics_track_id is set (as example) using .env file. 

**Dockerfile**, **automatic test**, and **documentation** in readme.md file was updated to handle this new feature.

* **Other information**:


kamil-kielczewski:
@SherlockStd I make this functionality that You want in #1736 : .env file is load in browser (not in webpack-building step) so you can give 'dist' folder to your client and change .env file inside it.

src/app/app.component.ts
```diff
@@ -64,13 +66,23 @@ export class AppComponent implements OnInit {
...
 
   constructor(
-    public appState: AppState
+    public appState: AppState,
+    public appEnv: AppEnv
   ) {}
 
   public ngOnInit() {
     console.log('Initial App State', this.appState.state);
+    this.initGoogleAnalytics();
```
gdi2290: this is easier to manage but it's better for these types of services to live outside so they're setup asap

kamil-kielczewski: I dont know do I understand you. Can you explain a little bit more what do you mean when you talk about 'these types of services" and "live outside"? Do you speak about 'initGoogleAnalytics' or 'AppEnv'? What exactly do you expect?  Please be concrete and clear - this will allow me to make proper fix.

almothafar: He means the starter meant to be at the basic level, as this line that he comments on is referring to another service that might not be needed as a "starter" if I'm starting project, the Google Analytics should be added in the last phase of my project not as a starting point that will waste sometime just to "start"! in short words, starter should be startup and run without pre-configuration.

kamil-kielczewski: Im not 100% sure that I understand you @almothafar - hoverer currently in code (on master - not in this pull request) googleAnalytics is invoked "at the top" of html file (so even before angular starts):
https://github.com/AngularClass/angular-starter/blob/master/src/index.html

The goal of this pull request was not changing (or fixing) this initGA behaviour but introduce support for .env files. I think better place to fix this separate topic will be another pull request - however I can change it here if you really want but tell me exactly where in code you wan to initialize GA (give fileName-line coordinates) because I not think about this problem and Im not sure - ?

almothafar: @kamil-kielczewski the solution is really simple, just comment out this line and keep it, so who wants it he can remove the comment mark and enable it, the code in index.html is not going to break anything.

And yes indeed you introducing support for .env files so no need to this line at all, it belongs to another pull request or feature.

I prefer to delete this line as what I said in lastly.

kamil-kielczewski: ok, if @gdi2290 approve your proposition I will do it (because in fact comment-out change pre-pull behaviour).

SherlockStd:
@kamil-kielczewski really nice thank you (sorry for the 1 month + delay ahah, wasn't active) ! :+1:

tsukhu:
Should we be considering the dotenv module https://www.npmjs.com/package/dotenv instead or are we adding something additional here ?

almothafar:
@tsukhu in some use cases we need to load these configs from JSON or .evn files, it is really helpful for different environments so you can load different config files just from the command line instead of editing javascript file for each environment.

In my version in my local machine I load them from JSON, in this PR from .evn file.

kamil-kielczewski:
@almothafar as far as I know dotenv works only for nodejs - however this pull request works at runtime in a browser (where dotenv doesn't works). Today or tommorow I will resolve conflicts - however this pull request was made 3 October and still no reaction from repository owner... :(

almothafar:
@kamil-kielczewski, when you do AoT build everything should be compiled and added to js files, so .env or JSON or whatever will be added to js files, this is where the issue #1902 talking about, maybe you need to try Angular v5.1 to see if your PR will still work or not, because AoT is default now in Angular v5.

kamil-kielczewski:
This pull request works with AoT. Old version where "environment file" was read on build time was here  #1736. However in this pull request due to [suggestion](https://github.com/gdi2290/angular-starter/pull/1736#issuecomment-302416421) made by @SherlockStd  I improve old solution to allow read env files WITHOUT COMPILATION - this means that you can compile application once and send compiled version to your clients and they only add .evn file to dist folder with their particular envirnment settings (so you can switch env files without changing application code and rebuild application)

This functionality is described with [The Twelve-Factor App Methodology](https://12factor.net/config):

> Env vars are easy to change between deploys without changing any code;

Which was more clear described in [this article](https://www.jvandemo.com/how-to-configure-your-angularjs-application-using-environment-variables/).

almothafar:
@kamil-kielczewski Lovely, this sounds awesome!

kamil-kielczewski:
Ok so I update this pull request. Everything works with aot, and we have here:
- .env files are load in browser at runtime (so we can exchnge it without code recompilation)
- environment configruation file `./.env` is copied at build step to `/dist/.env` - this paths are easy to remember and manipulate by third persons
- .env file should be NOT commited to repository (because its content is **envirnonment dependent**) - only .env.example file should be in repository
- in .env houd be only envirnoment dependent variables. Non environment dependent config-variables should be stored in different place (for instance in some .ts file compiled in build step)
- .env.exmaple file is used in automatic tests

Original pull request was made in 11 May, its new version in 3 October - now is 27 December -
 so more than 7.5 months later :(  - so I hope response for this pull request will be quicker. This message I address to: @gdi2290. And have question to @almothafar - to you have some abilities or ideas how to speed up process of accepting pull requests? This feature is really very basic and important in daily programmers life...

lanovoy:
This is interesting but seems overcomplicated... There is a much easier approach: create a json configuration environment specific: config.dev.json, config.staging.json, config.prod.json, provide environment into container when it's getting started (command line, docker-compose, stack whatever). Create a bash script that will be an entry point which should copy environment specific file to the dist folder without environment part (so the name is always the same): config.json. Finally, use APP_INITIALIZER provider in angular to pre-load config.json when app is getting bootstrapped. This way you can have full power of json. We've even added a webpack plugin which combines config.base.json with some common settings/defaults with environment specific one.

kamil-kielczewski:
@lanovoy - I don't sure that I full understand you - do you mean that is better to load .json environment config file instead .env environment config file? 

If only this is the problem that i implement parser of .env file (which is in fact quite easy and short part of code) I due it because [The Twelve-Factor](https://12factor.net/config) methodology recommend to use this  kind of os/language agnostic env variables - and also[ Laravel framework use this approach with success](https://laravel.com/docs/5.4/configuration#environment-configuration). I also wan't to say that in this PULL request I NOT USE ANY LIBRARY(!!! -  why you mention about that?). 

Last thing - env variables should be simple (not like json wich allow to do some nested things etc) - and there are NOT config files - e.g. in Laravel you have .env files and separately config .php files (which use sometimes env variables). So .env files ARE NOT replacement for CONFIG files - and this two kind of files are different and they are use for different purposes.

lanovoy:
You do not use libraries, but your code contains this complex method to parse env files - parseEnvToObject. Why do you need all that if you can just serve json? In any case neither your or mine approach fits the purpose of this repo... By far not every project out there requires environment specific settings.

lanovoy:
This is what I was talking about (still I don't think it should be part of this starter project) http://www.cicoria.com/ensuring-initializing-and-server-side-configuration-with-angular/

kamil-kielczewski:
A lot of projects need:
- development environment
- production environment

man also use test environment (and for instance: each of them use different backend API...).

Why to use .env files - they are very easy, system/language agnostic, they allow to use comments - which is very helpful for your client - you provide compiled application in dist folder with .env.example file, client copy it into .env file (because that fille should not be commitet into repo directly - is .gitignored) and edit by read instruction in comments - in json you cannot use comments.

kamil-kielczewski:
One more thing to clear: .env files SHOUD NOT be commited into repository.

lanovoy:
But you have to parse manually your env files. JSON is standard natively supported on the other hand. You can easily combine JSONs, parser is already there. 
I agree that many projects need environment specific settings, but not all. Of course it's up to author of this repo, but I'd keep your changes in a PR, so people who like this approach and who feel as they need it for their projects can easily grab it from your branch/fork.

kamil-kielczewski:
@lanovoy I have question because after my last comment I see "kamil-kielczewski closed this 10 minutes ago" - and this pull request change status to 'closed' -  but I don't "colose" anything (may be I do something accidently). How I cane reopen it again?

stormit-vn:
We are calling an HTTP request to retrieve environment variable at runtime, so someone can easy to get **all application configuration** variables (sensitive) information. I think it doesn't a good approach. Is it?

I am using the .env file to store environment variables. Basically, it works during development. But when building AOT release, all variables not combined for when initializing a module at the root level, but it works when we get value from an injected service. I would like to know the root cause so that we can find a solution to fix this issue instead of using this fix. I am using dotenv-safe and DefinePlugin to define variables

```
// Load environment variables using dotenv-safe to apply https://12factor.net/ approach
const envVars = require('dotenv-safe').load({
    path: options.env === 'development' ? '.env' : `.env.${options.env}`,
    allowEmptyValues: true,

}).parsed;
const envMap = _.mapValues(envVars, item => JSON.stringify(item));
```

```
new DefinePlugin({
    'HMR': METADATA.HMR,
    'AOT': METADATA.AOT,
    'process.ENV': JSON.stringify(METADATA.ENV),
    'process.NODE_ENV': JSON.stringify(METADATA.ENV),
    'process.HMR': JSON.stringify(METADATA.ENV),
    'APP_CONFIG': envMap
}),
```

and here is how I use the variables to configure module at the root level
```
const oktaAuthConfig = {
    clientId: APP_CONFIG.OKTA_CLIENT_ID,
    issuer: `${APP_CONFIG.OKTA_BASE_URL}/oauth2/default`,
    redirectUri: APP_CONFIG.OKTA_REDIRECT_URL,
    scope: 'openid profile address email phone',
    onAuthRequired: onOktaAuthRequired
};


imports: [
    OktaAuthModule.initAuth(oktaAuthConfig),
],
```

kamil-kielczewski:
> get all application configuration variables (sensitive) information.

It is not the case in frontend application - because assumption for this configiration varaibles is that they are PUBLIC.  If they are not stored in separate file like .env then they are stored in code directly which does not provide any protection against their reading. This pull request is not about protect .env variables (which are public by desing) but tu allow to easily echange them without application recompilation.

Of course instead using .env files we can also use some .json file or even .js file (which in fact simplify a little my code in this pull request).

kamil-kielczewski:
> get all application configuration variables (sensitive) information.

It is not the case in frontend application - because assumption for this configiration varaibles is that they are PUBLIC.  If they are not stored in separate file like .env then they are stored in code directly which does not provide any protection against their reading. This pull request is not about protect .env variables (which are public by desing) but tu allow to easily echange them without application recompilation.

Of course instead using .env files we can also use some .json file or even .js file (which in fact simplify a little my code in this pull request).

Nearnotes:
hello dude do sime example for vpn src.. Just to make a practice to make vpn app.. Thanks

kamil-kielczewski:
@Nearnotes  vpn src?

Nearnotes:
Src of vpn layout i am usibg aide for android

PR #1960, opened 2018-01-21
Some recommended improvements

kgish:
Refactored/simplified the dockerfile so that the run statement is optimized and the resulting docker image is much smaller. Also removed the bash and openssl versions as this causes problems.


lanovoy:
There is a better (recommended) way to do it. Called 'multi-staged builds'. See here https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#use-multi-stage-builds

lanovoy:
Dockerfile was updated to support multi-staged build and always create smallest possible images. If you have an ideas on how to improve it further please create a new PR.

PR #2019, opened 2018-06-23
Migration to Angular 6, RxJs 6 and Webpack 4

lanovoy:
Update of configuration and dependencies

Current version is based on Webpack 3, Jasmine 2, RxJs 5.5 and Angular 5.

- Migrated to Webpack 4 with optimized configs
- Got rid of outdated ngcWebpack (it's pretty much deprecated by the author) - using directly ngtools/webpack
- Migrated to Angular 6 and RxJs 6.2
- Added TSLint rules for RxJs 6.2
- Improved Karma config for local development - should now work similar to Angular/cli based projects
- Migrated to Jasmine 3


govi2010:
looks good to me.

lanovoy:
@gdi2290 when you get a chance please review this PR. A big step forward with this PR, so need more people to look into changes here before we merge it.

envil:
Looking forward for this to be merged.

briosheje:
Not sure whether I'm the only one experiencing the issue, but updating from the branch to my project, angular seems to be unable to resolve any css whatsoever, neither the ones package-related, nor the inline ones (or included ones).

Was that tested already or am I the only one experiencing that?

lanovoy:
@briosheje Maybe you merged something incorrectly? I have the same changes in our production code already and everything seem to be fine. You need to make sure path for mini-css-extract-plugin is set up correctly in your branch (webpack.prod.js)

briosheje:
@lanovoy No, sorry, everything was merged properly, the issue was slightly thinner: any css file should be included in `styles.scss`, I'm using some third-party libraries which were included in `polyfills.browsers.ts` (official guides explicitly recommended to include these files in *.ts files instead of .scss ones), moving them to scss imports apparently solved the issue, I'm only experiencing some strange icon sizes, but it's not related to the PR itself.

Perhaps it's worth mentioning that with this commit you won't be able to import .scss files in .ts files anymore, or, at least, that it may not work exactly as expected.

lanovoy:
Hmm, that's worth investigating more, since future versions of libSass will disallow importing css files using '@import'

briosheje:
@lanovoy the easiest you can try is importing twitter bootstrap, in my case version 4:

package.json:

`"bootstrap": "^4.0.0",`

If you import through `styles.css`:

`@import '~bootstrap/dist/css/bootstrap.min.css';`

It properly works.
If you try to import it through polyfills.browser.ts, it doesn't work:

`import 'bootstrap/dist/css/bootstrap.min.css';`

It does not throw any error or whatever, it just silently fails with no error (the css seems not to be bundled at all to me, but I might be wrong).

I'm not sure how to investigate further on that topic, I've always had issues with webpack's css resolving with electron some time ago, and literally gave up after hours of guessing. In that case, however, I had no issues at all (until angular 6, of course).

lanovoy:
@briosheje Did you try to import from app.module or main? polyfill.browser.ts is a separate entry point, so I'm not exactly sure if importing css from it and from the main would be effectively the same.
Also is it only a problem with production builds or is it the same with dev builds?

If you will have time, can you check if projects generated with latest angular cli have a same problem with importing css from ts files?

briosheje:
Sure, I quel try tomorrow morning.

Currently, importing from the css, i have some strange outputs in my production build, while in dev mode it’s properly working.

I will try to import everything in the main file and check both dev and prod builds

Ottieni Outlook per iOS<https://aka.ms/o0ukef>
________________________________

briosheje:
@lanovoy sorry for the late answer, but I had to test some stuff before being able to go on with gfx tests.

Apparently, external libraries does NOT work if they are included either in app.module or in main.browser.

Local scss and css files are fine, but whenever a .scss or css points to anything under node_modules I'm unable to make it work in devmode (not sure in prod mode, but I'm not going to test it).

For instance, I'm including "regular" local .scss and css files in my app.module, while I'm importing third party libraries through scss `@import` in some .scss files, though it's probably not the best practice.

I will keep you updated if I'm able to find any reason or any workaround to make it work through .ts

jurisz:
When I try to change target output from es5  to es2015 in tsconfig.webpack.json.
I got:
error TS1323: Dynamic import cannot be used when targeting ECMAScript 2015 modules.

briosheje:
@lanovoy Just to keep you updated, both dev and prod mode seems to somehow work by importing css in .scss files, regardless whether they are referenced locally (local css files) or npm dependencies.

I'm still confused about some strange css behaviors in devmode, while in prod mode they seems to be working fine.

I think this pull should just be release asap regardless the .scss issues, perhaps in the near future we may find a way to import .scss directly from .ts files instead (despite there still are so many libraries recommending to import through .scss).

What is interesting enough, and I think this should be mentioned, importing fontawesome now properly works both in dev and prod mode, in my case I'm following this approach in `styles.scss`:

```
$fa-font-path: "~font-awesome/fonts" !default;
@import "~font-awesome/scss/font-awesome";
```

and it's working beautifully, while in the previous version (webpack 3 and angular 5) this was **not** working in prod mode, so some magic happened there :).

config/webpack.common.js
```diff
@@ -26,19 +24,19 @@ const buildUtils = require('./build-utils');
...
+  const METADATA = Object.assign({}, buildUtils.DEFAULT_METADATA, options.metadata || {});
   const GTM_API_KEY = process.env.GTM_API_KEY || APP_CONFIG.gtmKey;
 
   const ngcWebpackConfig = buildUtils.ngcWebpackSetup(isProd, METADATA);
   const supportES2015 = buildUtils.supportES2015(METADATA.tsConfigPath);
 
   const entry = {
     polyfills: './src/polyfills.browser.ts',
-    main:      './src/main.browser.ts'
+    main: './src/main.browser.ts'
```
zzs1020: shouldn't these 2 paths start with '../src/' because webpack.common.js is in config folder?

config/webpack.common.js
```diff
@@ -26,19 +24,19 @@ const buildUtils = require('./build-utils');
...
-module.exports = function (options) {
+module.exports = function(options) {
   const isProd = options.env === 'production';
   const APP_CONFIG = require(process.env.ANGULAR_CONF_FILE || (isProd ? './config.prod.json' : './config.dev.json'));
 
-  const METADATA = Object.assign({}, buildUtils.DEFAULT_METADATA,options.metadata || {});
+  const METADATA = Object.assign({}, buildUtils.DEFAULT_METADATA, options.metadata || {});
   const GTM_API_KEY = process.env.GTM_API_KEY || APP_CONFIG.gtmKey;
 
   const ngcWebpackConfig = buildUtils.ngcWebpackSetup(isProd, METADATA);
```
zzs1020: why don't you delete this line? I thought you replaced ngcwebpack with ngtoolswebpack

config/webpack.common.js
```diff
@@ -26,19 +24,19 @@ const buildUtils = require('./build-utils');
...
   const supportES2015 = buildUtils.supportES2015(METADATA.tsConfigPath);
 
   const entry = {
     polyfills: './src/polyfills.browser.ts',
-    main:      './src/main.browser.ts'
+    main: './src/main.browser.ts'
   };
 
   Object.assign(ngcWebpackConfig.plugin, {
```
zzs1020: same here?

config/webpack.common.js
lanovoy: The config is still used, though. ngcWebpack is just a wrapper around webpack/ngtools and almost all config values are the same.

config/webpack.common.js
lanovoy: I suppose the path is relative to the folder in which node process for webpack is getting launched from, so that would be the root with package.json - not /config...

config/webpack.common.js
lanovoy: We might want to do some renaming here for clarity, though...

zzs1020: thanks!

PR #2024, opened 2018-07-16
missing && in package.json build command

vincent-cm:

Bug Fix

Build AoT hangs at building module: 69% building modules 1404/1405 modules 1 active …es/@firebase/database/dist/index.cjs.js

Build success

* **Other information**:


envil:
Why do you do that? I tried this and the the build is no longer AOT.

vincent-cm:
@envil Sorry about that I thought it was the && problem to make my build memory leak. Now I understand is the AngularFire2.

PR #2029, opened 2018-07-25
Fix file syncing between host and docker container

envil:
Bug fix

When we run `docker-compose up` and start developing, the files aren't synced to the docker container => no update nor rebuild triggered.

`docker-compose up` now sync file between host and container

* **Other information**:
- Added solution of node-sass environment changed (since when we sync the working folder, the node_modules folder (if exist) will override the one in docker). This cause the node environment changed and will throw a build error from node-sass. I added the commands to rebuild the node-sass.

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -12,9 +12,9 @@
 FROM node:8.9.4-alpine as builder
 
 # Uncomment this if you encounter node-sass environment changed issue (1/2)
-RUN apk add --no-cache python &&\
-    apk add --no-cache make &&\
-    apk add --no-cache g++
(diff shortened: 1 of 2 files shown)


envil:
@sabousha that's a good idea. I'm applying that to my running project to see if any problem.

sabousha:
Can we talk about when is it a good idea to include an image in the docker-compose file? I don't like having to build my docker image everytime I want to bring my server down due to a change in package json.  I will admit, this is only a problem when playing with dependencies, which do not change often.

envil:
@sabousha IMHO, there're no ways to avoid putting the server down in case of dependencies changed.

